PR #11, opened 2015-06-22
sasl-mechs.md: Add ECDSA-NIST256P-CHALLENGE

nyuszika7h:
~~Also, put PLAIN first, because it's the most basic mechanism.~~ I might do that in a separate pull request.

### DIFF
--- a/docs/sasl-mechs.md
+++ b/docs/sasl-mechs.md
@@ -6,6 +6,7 @@ IAL primarily uses the same mechanisms as SASL in other protocols. Most commonly
 
 * [EXTERNAL](https://tools.ietf.org/html/rfc4422#appendix-A) as defined by RFC 4422
 * [PLAIN](https://tools.ietf.org/search/rfc4616) as defined by RFC 4616
+* [ECDSA-NIST256P-CHALLENGE](https://github.com/kaniini/ecdsatool)
 * [SCRAM-SHA-256](https://tools.ietf.org/html/draft-hansen-scram-sha256-02)
 
 as well as custom ones:


docs/sasl-mechs.md
```diff
@@ -4,8 +4,9 @@ title: SASL Mechanisms
 ---
 IAL primarily uses the same mechanisms as SASL in other protocols. Most commonly:
 
-* [EXTERNAL](https://tools.ietf.org/html/rfc4422#appendix-A) as defined by RFC 4422
 * [PLAIN](https://tools.ietf.org/search/rfc4616) as defined by RFC 4616
+* [EXTERNAL](https://tools.ietf.org/html/rfc4422#appendix-A) as defined by RFC 4422
+* [ECDSA-NIST256P-CHALLENGE](https://github.com/kaniini/ecdsatool)
```
Mikaela: I think this link should go to https://github.com/kaniini/ecdsatool#mechanism-spec instead and there should be something more formal than it.

nyuszika7h: I don't think so, because you would have to scroll up anyway to see how to use ecdsatool.

DanielOaks:
I agree with putting PLAIN first. Seems to be the globally available one these days.

Should `ECDSA-NIST256P-CHALLENGE` be documented as a formal IRCv3 SASL mech ala [the others](https://github.com/ircv3/ircv3-specifications/tree/master/documentation)? I'm iffy about linking it to a specific tool's readme.

I've never done that much with SASL though, will probably leave for someone else to take a look at.

docs/sasl-mechs.md
Mikaela: But ecdsatool isn't related to the specification or do you see those RFCs including anything else either?

Mikaela:
> I agree with putting PLAIN first. Seems to be the most globally available / basic one these days.

I also think that PLAIN and EXTERNAL are good to have as first, I am not sure on Atheme, but with Anope they are hardcoded into the SASL module so you cannot disable them (this was wondered when ZNC removed DH-\* if I recall correctly).

> Should ECDSA-NIST256P-CHALLENGE be documented as a formal IRCv3 SASL mech ala the others? I'm iffy about linking it to a specific tool's readme.

:+1:

kaniini:
:-1:

nyuszika7h:
@kaniini You're not contributing anything to this discussion at all. Is it so hard to explain what you would prefer?

Anyway, I should probably separate the two changes out, though I'm not sure if moving PLAIN is worth a pull request on its own.

docs/sasl-mechs.md
nyuszika7h: @Mikaela True, you would probably use a library (or an FFI binding for `libecdsaauth` perhaps) unless you're writing a client in bash.

DanielOaks:
I can throw PLAIN up the top after this PR gets decided.

kaniini:
:-1: to including ECDSA-NIST256P-CHALLENGE at all.  And yes, I get to say that because I made the damned thing.

nyuszika7h:
@kaniini Even if you think it sucks (because of NIST or whatever), it's already being used out in the wild (e.g. freenode), so IMO if DH-AES and DH-BLOWFISH are still mentioned, then ECDSA-NIST256P-CHALLENGE should be listed too. But I haven't heard you or anyone saying that it shouldn't be used, so I don't understand what's the problem.

kaniini:
I don't think what I made sucks, but it is not an IRCv3 standard nor do I have any intentions of submitting it as an IRCv3 standard.  Instead, I plan on revising the standard and introducing a family of mechanisms based on various curve types (NIST, 25519, etc.) through an IETF submission.  But that will require more time than I have to work on it right now.

kaniini:
For what it's worth, documenting DH-AES and DH-BLOWFISH here *is* a mistake because it implies endorsement of those mechanisms, and they do legitimately suck.

nyuszika7h:
> it is not an IRCv3 standard

Nothing on that page implies those are all standard mechanisms. SCRAM-SHA-256 is just a draft as well.

>  Instead, I plan on revising the standard and introducing a family of mechanisms based on various curve types (NIST, 25519, etc.) through an IETF submission.

Nothing is submitted to IETF yet, is it? Using that logic, why not remove everything from the site because nothing is an IETF standard?

nyuszika7h:
> because it implies endorsement of those mechanisms

There's a big red warning.

kaniini:
I still think the page, in it's entirety, is inappropriate, and do not want my mechanism documented there.  Thank you for respecting my request.

PR #13, opened 2015-07-08
Add simple 404 page

### LINKED ISSUE #12: More friendly 404 page?
* https://help.github.com/articles/custom-404-pages/

The current GitHub pages 404 error isn't very helpful, I think it should at least link back to index.

Mikaela:
I was hoping to make it more helpful including useful links, but I have
no idea how they would be shown in a good way.

Closes #12

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,0 +1,7 @@
+---
+layout: page
+permalink: /404.html
+sitemap: false
+---
+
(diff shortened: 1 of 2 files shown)


DanielOaks:
Sounds good, we can expand upon it. Thanks!

PR #15, opened 2015-07-12
Minor html validation fixes

dequis:
Also changed slightly the style of `<warn>` to just use a bootstrap
alert-warning instead


DanielOaks:
Looks good, thanks!

PR #16, opened 2015-07-22
clients: Add mIRC (CAP, tls, server-time)

Aerdan:


kaniini:
I disagree with this as it doesn't natively support SASL, a requirement for IRCv3.1 compliance.

DanielOaks:
I'll leave this in for now and look at a better way to display nonconformant clients, as with the software in #17

Issue #17, opened 2015-07-22
Some software is listed in IRCv3 compliant software without even IRCv3.1 base compatibility


Mikaela:
The base requirements for 3.1 are cap, multi-prefix and sasl. The following software doesn't fill the requirements:

## Clients

* BitchX (multi-prefix)
* Konversation (multi-prefix, sasl)
* LimeChat (multi-prefix)
* mIRC (multi-prefix, sasl)
* Mibbit (multi-prefix, sasl)
* AndChat (multi-prefix)
* AndroIRC (multi-prefix)
* IRC7 (multi-prefix)
* LimeChat mobile (multi-prefix)

## Servers

* IRCCloud (sasl)
* mammon-ircd (multi-prefix, sasl)

## Libraries

* Net::Async::IRC (extended-join, multi-prefix)

DanielOaks:
Looking at a different way to display non-conformant software. Will go through some ideas and post once I've got something that looks decent

jwheare:
12:43 @Mikaela no access to github at the moment and my only complaint against not base compatible cöients is resolved by that PR (#35) when it gets merged

This issue can be closed.

PR #21, opened 2015-08-01
clients: Add AdiIRC.

Aerdan:


Mikaela:
:-1: as AdiIRC isn't IRCv3.1 base compatible by missing SASL. I think it's better to leave not-compliant software out until https://github.com/ircv3/ircv3.github.io/issues/17#issuecomment-123927921 is implemented.

DanielOaks:
Given that this doesn't support 3.1 core, I'll leave this out 'til we've
got the way to show non-compliant software merged.

kaniini:
Isn't the whole point to show compliant software in order to encourage developers to participate in the compliance scheme?

DanielOaks:
I think listing the software, but prominently negatively marking them would work better. Being able to display the status of IRCv3 implementation, even on software which isn't fully compliant, is useful because it lets people see the support for our specifications, and lets interested users see the software support across the board.

Clearly marking them negatively should dissuade people visiting the page from using them based on the link there. The main thing is to make sure we're not advertising the software that's not compliant, so I'm keeping that in mind while going through different options.

jwheare:
+1 for including as much software as possible on these lists. Compatibility is indicated in the matrix, inclusion by no means indicates endorsement. I personally would love to be able to see how far ahead different clients are with support, even if they only support one CAP. And yes, mIRC should be on this list, and not buried away in a forum where it can't be easily compared with other clients http://forums.mirc.com/ubbthreads.php/topics/244662/IRCv3

jwheare:
Oh my bad, mIRC is already on the list. Good!

jwheare:
There is consensus on IRC from anyone who has expressed an opinion and been active recently: @jwheare @dequis @Mikaela

attilamolnar:
Thanks!

PR #22, opened 2015-08-01
Libs: Communi 3.4 supports account-tag & chghost

jpnurmi:
http://communi.github.io/doc/3.4/ircv3.html


DanielOaks:
Awesome, thanks!

PR #23, opened 2015-08-08
clients: WeeChat supports MONITOR via /notify

maxteufel:


Mikaela:
Is there a reason why this isn't merged? Is comment from @FlashCode needed?

Git version also supports `cap-notify`, but it probably shouldn't be added before it's in stable release.

attilamolnar:
Thanks!

PR #24, opened 2015-08-11
software: ZNC 1.7 adds support for new caps

jpnurmi:
- cap ls 302
- account-notify
- away-notify
- extended-join
- cap-notify

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -6,12 +6,17 @@
         v3.1:
           cap: "0.094+ client→ZNC,<br/>0.090+ ZNC→server"
           multi-prefix: "0.094+ client→ZNC,<br/>0.090+ ZNC→server"
+          account-notify: "Git"
+          away-notify: "Git"
+          extended-join: "Git"
         v3.2:
           userhost-in-names: "0.094+ client→ZNC,<br/>0.090+ ZNC→server"
       partial:
         v3.1:
           sasl: "1.0+, ZNC→server only"
         v3.2:
+          cap: "Git, client→ZNC only"
+          cap-notify: "Git, client→ZNC only"
           echo-message: "Git, client→ZNC only"
           batch: "1.6.x as <code>znc.in/batch</code>, Git as <code>batch</code>, client→ZNC only"
           server-time: "1.2–1.6.x as <code>znc.in/server-time-iso</code>, Git as <code>server-time</code>, client→ZNC only"


DarthGandalf:
Please mark it as git. 1.7 is not released yet.

2015-08-11 16:21 GMT+01:00 J-P Nurmi <notifications@github.com>:

DarthGandalf:
cap-notify is supported only on client side, I thought?

2015-08-11 16:21 GMT+01:00 J-P Nurmi <notifications@github.com>:

jpnurmi:
> Please mark it as git. 1.7 is not released yet.

Done

> cap-notify is supported only on client side, I thought?

Right, thanks for the reminder. :)

Mikaela:
I don't think those caps should be mentioned until 1.7.0 is released. Listing only stable versions could be more user friendly.

DarthGandalf:
I thought this table is more for developers than end users?

Mikaela:
> I thought this table is more for developers than end users?

Possible, but I am linking people who want to know what are good IRC clients into it.

Mikaela:
Does 1.6.x mean that it's in that git branch or came in 1.6.x version? If the latter, I think it would be easier to understand if it was said that it came with 1.6.0 or 1.6.x.

DarthGandalf:
It means all versions from 1.6.0 to 1.6.[whatever will be the last one]

Mikaela:
In that case I think saying 1.6.0 is more helpful as with the x you cannot know if it was added in 1.6.1 and are still running 1.6.0.

DarthGandalf:
There are 2 usages of it:
1. batch: 1.6.x as `znc.in/batch`, Git as `batch`
2. server-time: 1.2–1.6.x as `znc.in/server-time-iso`, Git as `server-time`

For "1", 1.6.0+ would make sense, yes.
For "2", not so much. 1.2–1.6.0 is incorrect, because it's the same in 1.6.1. 1.2–1.6.1 would be right, but only until 1.6.2 is released. Do you want to update this table every time ZNC gets a new release in 1.6 series?

DarthGandalf:
Though... that one can be "1.2+"

Mikaela:
For 1 it would make a little more sense as some minor feature can come with 1.6.1. And for 2 as you said it could be 1.2+. With `1.2-1.6.x` it also sounds like it must be updated when 1.7 is released.

PR #26, opened 2015-08-22
contact.html: fix freenode address & use KiwiIRC

Mikaela:
I believe the previous format wasn't correct as it was missing the port
number and channel prefix. Missing SSL while it will be encouraged (STS)
also appears weird.

I replaced QWebIRC with Kiwi, because it doesn't support IRCv3 while
Kiwi does and Kiwi is also part of stakeholders.

### DIFF
--- a/contact.html
+++ b/contact.html
@@ -3,6 +3,6 @@ title: Contact
 layout: default
 ---
       <div class="alert alert-info">
-        <strong>Have your own IRC client?</strong> You can chat with us on <a href="irc://irc.freenode.net/ircv3"><strong>irc.freenode.net</strong></a> in <a href="irc://irc.freenode.net/ircv3"><strong>#ircv3</strong></a>.
+        <strong>Have your own IRC client?</strong> You can chat with us on <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>irc.freenode.net</strong></a> in <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>#ircv3</strong></a>.
       </div>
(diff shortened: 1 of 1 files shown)


contact.html
```diff
@@ -3,6 +3,6 @@
 layout: default
 ---
       <div class="alert alert-info">
-        <strong>Have your own IRC client?</strong> You can chat with us on <a href="irc://irc.freenode.net/ircv3"><strong>irc.freenode.net</strong></a> in <a href="irc://irc.freenode.net/ircv3"><strong>#ircv3</strong></a>.
+        <strong>Have your own IRC client?</strong> You can chat with us on <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>irc.freenode.net</strong></a> in <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>#ircv3</strong></a>.
```
Zarthus: Using a # is incorrect.

Mikaela: How so?

https://tools.ietf.org/html/draft-butcher-irc-url-04#section-2.5.1

Mikaela: Or do you mean I should ircs://irc.freenode.net:6697/%23ircv3 instead?

Mikaela: `2015-09-26 16:32:48+0300 < Zarthus> Mikaela: re ircv3: turns out you're more right than I was. I've been using irc://irc.network.net/channel forever since my client detects irc://irc.network.net/#channel as ##channel`

Mikaela:
Conflict fixed.

PR #27, opened 2015-08-30
Update capabilities support of Limnoria

ProgVal:
These capabilities were added in the last release.
`cap-notify` is partially supported: `CAP DEL` is properly handled, and `CAP ADD` is ignored

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -313,6 +313,11 @@
           - metadata
           - monitor
           - account-tag
+          - cap-notify: "CAP ADD is ignored"
+          - chghost
+          - invite-notify
+          - server-time
+          - userhost-in-names
     - name: Willie
       link: http://willie.dftba.net
       support:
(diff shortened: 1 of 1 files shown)


DanielOaks:
That `- cap-notify: "CAP ADD is ignored"` is ignored.

For what you want to do, you need to make `v3.2` into a dictionary so it can properly detect `cap-notify` as a key. Something like this:

```yaml
    - name: Limnoria
      link: https://github.com/ProgVal/Limnoria
      support:
        v3.1:
          - cap
          - multi-prefix
          - away-notify
          - account-notify
          - extended-join
          - sasl
        v3.2:
          cap:
          metadata:
          monitor:
          account-tag:
          cap-notify: "CAP ADD is ignored"
          chghost:
          invite-notify:
          server-time:
          userhost-in-names:
```

DarthGandalf:
By `CAP ADD` you mean `CAP NEW`?

ProgVal:
Indeed, I meant `CAP NEW`.
I just fixed this and moved `cap-notify` to `partial`.

DanielOaks:
Looks good, thanks.

PR #31, opened 2015-10-20
so_clients.yml: add WeeChat version numbers

Mikaela:
I think seeing when different capabilities were added at glance can be
interesting.

Source: https://weechat.org/files/changelog/ChangeLog-devel.html

I also changed WeeChat homepage to https as I think it's not enough big
change to deserve it's own commit.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -135,19 +135,19 @@
           - sasl
           - userhost-in-names
     - name: WeeChat
-      link: http://www.weechat.org
+      link: https://weechat.org
       support:
         v3.1:
-          - cap
-          - account-notify
-          - away-notify
-          - extended-join
-          - multi-prefix
-          - sasl
+          cap: 0.3.2+
+          account-notify: 1.2+
+          away-notify: 1.0+
+          extended-join: 1.1+
+          multi-prefix: 0.0.1+
+          sasl: 0.3.2+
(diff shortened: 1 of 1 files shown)


Mikaela:
The update is that I changed cap to say that it was added at the same time as SASL. I am not fully sure if this is true, but the original [WeeChat task on SASL support at Savannah](https://savannah.nongnu.org/task/?8829) hints that WeeChat didn't support cap before SASL.

Time when multi-prefix support was added is unknown as it doesn't read in changelog or git commits. ZNC wiki claims it was added in 0.3.7 which was when option for requesting capabilities from server was added. However @FlashCode has said that it has been there for a long time, at least 0.3.0 for manual `/quote cap req`. I think it's the best that I leave it as empty/yes for now, but in case you prefer either of the before mentioned version numbers or something else, please let me now and I will change it.

Mikaela:
```
2015-10-20 12:56:00+0300 <@FlashCode> Mikaela: you could also remove the www. in www.weechat.org ;)
2015-10-20 12:56:07+0300 <@FlashCode> (even if it works, it's deprecated)
2015-10-20 12:57:23+0300 < Mikaela>  done
```

Mikaela:
@FlashCode checked that WeeChat has supported multi-prefix since 0.0.1 which was the first public release of WeeChat.

PR #34, opened 2015-11-04
Fix/clarify IRCCloud server support

jwheare:
* The server requires TLS and PASS for auth, so sasl, tls, account-notify caps are N/A. 
* Fixed userhost-in-names in 3.2 section.


Mikaela:
Could there be another template like "partial", but named as "not applicable" or something and be gray instead of yellow?

I don't think this PR is the only case where it's needed, for example some bot might not use/need MONITOR for anything so it doesn't need to support it and this was nearby the case with Limnoria until one case where it could be useful was pointed out.

dequis:
:+1:

jwheare:
Squashed to a single commit. @Mikaela's suggestion of a separate "not applicable" styling was implemented.

Issue #20, opened 2015-07-22
Software: Specify OS support for clients


Mikaela:
At #znc it was asked if there is any good Android IRC app which works well with ZNC and I replied:

> any from http://ircv3.net/software/clients.html that supports server-time, I would say
> it doesn't separate by mobile OS though
> none it looks like

I think that especially with mobile the OS should be indicated somehow, but it wouldn't hurt with others either, it just must be kept in mind that clients can support multiple OSes.

attilamolnar:
I don't think we should do this. The IRCv3 site is not an IRC help site and is certainly not a client guide for users, it's mainly aimed at developers.

Mikaela:
There is already list of clients and what capabilities they support which probably interests more IRCv3-enthustiastic users than developers and is also the place where people link when someone asks for good IRC client. If the list didn't exist, users are discouraged from using IRCv3 capable clients and client is easier to find when you know one for your OS.

attilamolnar:
Implemented by @DanielOaks in #36.

PR #36, opened 2015-11-10
software: Add OS icons for mobile clients

DanielOaks:
Adds OS icons for mobile clients, as #20 

Looks like this:
Mobile Clients page

### DIFF
--- a/_data/os.yml
+++ b/_data/os.yml
@@ -1,0 +1,12 @@
+# mobile
+android:
+  name: Android
+  icon: android
+
+ios:
+  name: iOS
+  icon: apple
+
+windows-phone:
+  name: Windows Phone
+  icon: windows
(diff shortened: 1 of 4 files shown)


jwheare:
Looks great

Mikaela:
:+1: something like that would also be nice for desktop clients, but mobile probably does need it more.

Mikaela:
Or wait, how do you separate OS X from iOS and Windows Mobile from Windows (when/if my previous comment happens)?

Just taking a look at mobile clients versus desktop clients or is there a separate logo availble for it?

DanielOaks:
No separate logo available, code-wise it's named differently so it'll have a different hover-over name, but I think being in the Desktop/Mobile categories will separate them enough. We can look at desktop clients after mobile ones get it, though.

dequis:
:+1:, i think the title="..." attribute (mouseover/tooltip text) is enough

attilamolnar:
LGTM

attilamolnar:
Thanks!

PR #37, opened 2015-11-10
Add more specific and useful meta descriptions for pages, fixes #7

### LINKED ISSUE #7: Pages could have meta descriptions
* https://encrypted.google.com/search?q=site:ircv3.net

That search doesn't tell a lot about what IRCv3 is and most of meta descriptions are just `IRCv3 development community.`

They can be added by adding `excerpt: "meta description here"` in the YAML frontend matter or whatever it was called as.

DanielOaks:
This just adds more useful meta descriptions for pages, which fixes #7 and should makes our search engine results display more nicely.

### DIFF
--- a/contact.html
+++ b/contact.html
@@ -1,6 +1,7 @@
 ---
 title: Contact
 layout: default
+meta-description: Feel free to talk with and participate in the IRCv3 Working Group on irc.freenode.net/#ircv3
 ---
       <div class="alert alert-info">
         <strong>Have your own IRC client?</strong> You can chat with us on <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>irc.freenode.net</strong></a> in <a href="ircs://irc.freenode.net:6697/#ircv3"><strong>#ircv3</strong></a>.
--- a/404.md
+++ b/404.md
@@ -3,6 +3,7 @@ layout: default
 permalink: /404.html
 sitemap: false
 title: 404 - Page Not Found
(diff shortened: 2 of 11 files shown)


contact.html
```diff
@@ -1,6 +1,7 @@
 ---
 title: Contact
 layout: default
+meta-description: Feel free to talk with and participate in the IRCv3 Working Group on irc.freenode.net/#ircv3
```
Mikaela: I think that maybe this should have the protocol and port, but as it's meta description, I am not so sure.

Mikaela:
:+1: except that some of the descriptions are a little short, but I have nothing to add to them and that can probably be done later as people invent more  descriptions.

DanielOaks:
They need to be a bit short to fit on the Google/etc page, since they give us a limited space and we want to fit inside it. We can always go over them some more though, see if we can make them clearer or present some more useful info.

Mikaela:
It's just that Google Webmaster tools (does anyone use them with ircv3.net) complain if something is too short or doesn't say anything useful.

contact.html
DanielOaks: Yeah, I was thinking about that but since it's just a meta description (and likely no clickable links), I think leaving it at this should be fine. Hopefully most of the people checking us out know Freenode (or just IRC in general) well enough anyways.

attilamolnar:
Thanks!

PR #38, opened 2015-11-10
Add .travis.yml for testing HTML5 validity

Mikaela:
I had this branch in my fork and I thought why to not open a PR in case this is found useful after all. I think I might have mentioned this once at IRC, but I don't remember if anything was said.

What it does is installing github-pages environment, building the site, installing offline HTML5 validator and running it.

If I recall correctly the site doesn't pass and this most likely isn't interesting at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: ruby
+sudo: false
+script:
+  - gem install github-pages
+  - jekyll build --drafts
+  - python2 --version && python2 -m pip --version
+  - python2 -m pip install html5validator --user --upgrade
+  - ~/.local/bin/html5validator --root _site/
+rvm: 2.1
+notifications:
+    email: false
+matrix:
+    fast_finish: true
+env:
+  global:
+  - NOKOGIRI_USE_SYSTEM_LIBRARIES=true


DanielOaks:
I could possibly see this being useful for running checks against PRs and making sure they still build, but I don't think we need the validator run on every commit. Running the validator here locally it's given me some interesting stuff to look at regarding the manual HTML templates, but I think putting that our CI file is overkill.

Will keep an eye on this and do some tests with it in my fork.

dequis:
I think the main thing we'd want to validate is the YAML, since it's easy to get the data structures subtly wrong. If doing the jekyll build takes care of that, running this would help. Although there are some mistakes that it won't catch (like lists that are `[1, 2, {3: "value"}, 4]`, valid but the templates will just skip the 3)

jwheare:
An HTML validator would have caught the unencoded ampersand here: 9fb898f so I'd say it would be useful, after some initial cleanup.

DanielOaks:
Fair, I'll do some HTML cleanups and then look at integrating the HTML5 validator. And that does sound useful @dequis, hopefully shouldn't be too hard to write a little script to load and check them.

attilamolnar:
I think it's useful to have, is there anything to lose?

DanielOaks:
Nothing to lose, but at the moment I'm going through and fixing up the site to not throw a zillion errors because of some of our static HTML / generation. I'll write up a YAML file checker and then merge this in.

Mikaela:
> An HTML validator would have caught the unencoded ampersand here: 9fb898f so I'd say it would be useful, after some initial cleanup.

There is just issue that while it's invalid HTML, IRC clients don't like valid one.

DanielOaks:
Because it's the `http` link that had the invalid character, browser/site would handle the valid one in the right way (since that link's not going to an IRC client).

DanielOaks:
Thanks for the work!

PR #39, opened 2015-11-10
Software: re-verified all caps and updated website links

dequis:
A different pull request will include the reference URLs (to make future updates easier)


DanielOaks:
Cool, thanks for the work

PR #42, opened 2015-11-11
Software: add reference URLs in comments to make future updates easier

dequis:


attilamolnar:
More links for inspircd:

https://github.com/inspircd/inspircd-extras/blob/master/2.0/m_capnotify.cpp
https://github.com/inspircd/inspircd-extras/blob/master/2.0/m_invitenotify.cpp
https://github.com/inspircd/inspircd-extras/blob/master/2.0/m_replaymsg.cpp

dequis:
Added a search for "GenericCap" in that repo too. Trying to keep it more or less generic to avoid the need to update URLs if files are moved in the repositories (although some entries are very specific but using a named tag / commit hash)

DanielOaks:
LGTM, we can always clean things up and add more info as we go.

DanielOaks:
Awesome, thanks.

PR #44, opened 2015-11-12
specs: Add a big alert on WiP (IRCv3.3) specification pages

DanielOaks:
This adds a big alert at the top of IRCv3.3 specs. This is good because it makes sure people see that they are not complete.

Looks like:
Alert Box


Mikaela:
:+1:

attilamolnar:
:+1:

PR #48, opened 2015-12-17
Add SASL mechanisms to the software table.

SaberUK:
I'm abusing the required attribute to show the ones that people "should" be implementing. If anyone has a better idea for this then I can change it.

### DIFF
--- a/_data/irc_versions.yml
+++ b/_data/irc_versions.yml
@@ -91,3 +91,39 @@ v3.2:
       name: userhost-in-names
       description: userhost-in-names Extension
       link: http://ircv3.net/specs/extensions/userhost-in-names-3.2.html
+
+SASL:
+  name: SASL Mechanism
+  specs:
+    external:
+      name: EXTERNAL
+      description: EXTERNAL SASL mechanism
+      required: true
+      link: "https://tools.ietf.org/html/rfc4422#appendix-A"
+    plain:
+      name: PLAIN
+      description: PLAIN SASL mechanism
+      required: true
+      link: https://tools.ietf.org/search/rfc4616
+    scram-sha-256:
+      name: SCRAM-SHA-256
+      description: SASL SCRAM-SHA-256 mechanism
+      required: true
+      link: https://tools.ietf.org/html/draft-hansen-scram-sha256-02
+    authcookie:
+      name: AUTHCOOKIE
+      description: Atheme AUTHCOOKIE SASL mechanism
+      link: https://github.com/atheme/atheme/blob/master/modules/saslserv/authcookie.c
+    dh-aes:
+      name: DH-AES
+      description: DH-AES SASL mechanism (deprecated)
+      link: http://ircv3.net/specs/documentation/sasl-dh-aes.html
+    dh-blowfish:
+      name: DH-BLOWFISH
+      description: DH-BLOWFISH SASL mechanism (deprecated)
+      link: http://ircv3.net/specs/documentation/sasl-dh-blowfish.html
+    ecdsa-nist256p-challenge:
+      name: ECDSA-NIST256P-CHALLENGE
+      description: Atheme ECDSA-NIST256P-CHALLENGE SASL mechanism
+      link: https://github.com/kaniini/ecdsatool
+
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -15,6 +15,9 @@
         v3.2:
           server-time: 1.9.7
           userhost-in-names:
+        SASL:
+          # AdiIRC is closed source so this is a guess.
+          - plain
     - name: BitchX
       # ref: https://github.com/BitchX/BitchX/search?q=%22CAP+REQ%22&
       link: http://www.bitchx.com
@@ -22,6 +25,8 @@
         v3.1:
           - cap
           - sasl
+        SASL:
+          - plain
     - name: Colloquy
       # ref: handleCapWithParameters() in https://github.com/colloquy/colloquy/blob/master/Chat%20Core/MVIRCChatConnection.m
       link: http://www.colloquy.info
@@ -48,6 +53,8 @@
           - userhost-in-names
         v3.3:
           - intents
+        SASL:
+          - plain
     - name: Conspire
       # ref: https://github.com/kaniini/conspire/search?q=cap_add_cap
       link: https://github.com/kaniini/conspire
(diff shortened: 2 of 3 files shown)


grawity:
maybe there should be a more generic field "status: required/optional/deprecated/dear gods no"

DanielOaks:
This looks nice, and thanks for that `partial` fix.

I'm against showing DH-AES and DH-BLOWFISH on the site in the same way we're showing everything else. For that matter, given #11 I'm very iffy on including `ECDSA-NIST256P-CHALLENGE` on there as well.

If we display them at all, my initial thought is to add a 'deprecated' class to every cell in that column and make them very desaturated, darker grey versions of the green, and just a darker grey instead of red for not supported. And for the `title` attribute of those cells to be set to `DEPRECATED` to make it a bit more clear on hovering over them. Hell, maybe even instead of `No`, we just write `Deprecated` or something similar to make it very clear that we don't like those mechanisms and that not implementing them isn't a fault.

That sort of thing could definitely be done with some creative extension of the `irc_versions.yml` file. And I could always hack away at it 'til I get something that mostly works all else fails. But I think it would probably just be easier and cleaner to see whether that column is marked `deprecated` in `irc_versions.yml`, and if so then just hide it.

Regarding `SCRAM-SHA-256`, I don't know enough about crypto and SASL to take a proper look at the whole 'required' thing. It looks alright to me, but I'd want some other people to weigh in on whether or not it should be there.

I definitely really like the idea. I'm just a little iffy about including it right here, along with those `deprecated`/`non-standard`/etc concerns. If we do have this, maybe a `services` page would be in order sometime after this is merged into master?

grawity:
> Regarding SCRAM-SHA-256, I don't know enough about crypto and SASL to take a proper look at the whole 'required' thing. It looks alright to me, but I'd want some other people to weigh in on whether or not it should be there.

XMPP has chosen SCRAM-SHA-1 as "required" (replacing the previous DIGEST-MD5 requirement). As non-cryptographer, I think it looks good when properly implemented – unlike my earlier Atheme module. (That is, services must only store ServerKey and StoredKey, because PBKDF2(password) is password-equivalent.)

Mikaela:
> I'm against showing DH-AES and DH-BLOWFISH on the site in the same way we're showing everything else.

:+1:. Listing them makes it seem that all clients which removed support for them when they were declared insecure and removed wrong Atheme are doing something wrong. At least WeeChat, ZNC and HexChat did this.

> For that matter, given #11 I'm very iffy on including ECDSA-NIST256P-CHALLENGE on there as well.

I have no own opinion, but would assume that @kaniini doesn't want it to be included based on them not wanting to include it on SASL mechanisms page as seen in #11.

Otherwise I like the idea.

> I'm abusing the required attribute to show the ones that people "should" be implementing. 

I also ask in this thread, where else than in this PR the mechanisms that "should" be implemented are listed?

_data/sw_clients.yml
```diff
@@ -71,6 +80,11 @@
           - server-time
           - monitor
           - userhost-in-names
+        SASL:
+          - dh-aes
+          - dh-blowfish
```
Mikaela: Source for dh-aes and dh-blowfish being supported by HexChat? I see only `SASL (username + password)` and `SASL EXTERNAL (cert)` in the login methods and am quite sure that they were removed long time ago.

_data/irc_versions.yml
```diff
@@ -91,3 +91,39 @@ v3.2:
...
+      link: https://tools.ietf.org/search/rfc4616
+    scram-sha-256:
+      name: SCRAM-SHA-256
+      description: SASL SCRAM-SHA-256 mechanism
+      required: true
+      link: https://tools.ietf.org/html/draft-hansen-scram-sha256-02
+    authcookie:
+      name: AUTHCOOKIE
+      description: Atheme AUTHCOOKIE SASL mechanism
+      link: https://github.com/atheme/atheme/blob/master/modules/saslserv/authcookie.c
```
attilamolnar: This is a homegrown mechanism not standardized anywhere, it should not be here.

_data/irc_versions.yml
```diff
@@ -91,3 +91,39 @@ v3.2:
...
+    dh-aes:
+      name: DH-AES
+      description: DH-AES SASL mechanism (deprecated)
+      link: http://ircv3.net/specs/documentation/sasl-dh-aes.html
+    dh-blowfish:
+      name: DH-BLOWFISH
+      description: DH-BLOWFISH SASL mechanism (deprecated)
+      link: http://ircv3.net/specs/documentation/sasl-dh-blowfish.html
```
attilamolnar: I see no point in listing these if they're deprecated.

_data/sw_clients.yml
```diff
@@ -190,6 +223,12 @@
           server-time: 0.4.0+
           userhost-in-names: 0.4.1+
           cap-notify: git
+        SASL:
+          - dh-aes
+          - dh-blowfish
+          - ecdsa-nist256p-challenge
```
Mikaela: WeeChat doesn't support dh-aes or dh-blowfish, they were removed long time ago like other clients did.

_data/irc_versions.yml
DanielOaks: I'd agree with this. I can handle this with the restructuring stuff I'm working with on top of this PR.

_data/sw_clients.yml
```diff
@@ -149,6 +174,8 @@
           sasl: 19+
         v3.2:
           monitor: 15+
+        SASL:
+          - plain
```
Zarthus: mIRC supports SASL? I thought only by a whacky "third-party" script. There's no GUI for it as far as I know, and the third party script was something I personally never could get working (hence me piping my connection through ZNC)

DanielOaks:
> I also ask in this thread, where else than in this PR the mechanisms that "should" be implemented are listed?

The http://ircv3.net/docs/sasl-mechs.html page does already list these mechanisms specifically. Aside from that, I'm leaving this issue open for opinions on those required ones.

_data/sw_clients.yml
Mikaela: I am wrong.

> Option "irc.server_default.sasl_mechanism":<nr>
> description: mechanism for SASL authentication: "plain" for plain text password, "ecdsa-nist256p-challenge" for key-based challenge authentication, "external" for authentication using client side SSL cert, "dh-blowfish" for blowfish crypted password (insecure, not recommended), "dh-aes" for AES crypted password (insecure, not recommended)

_data/sw_clients.yml
Mikaela: Wrong again `< TingPing> Mikaela, it still does though it was planned to be removed`

DanielOaks:
Merged this into master manually, with the table now living on the [sasl-mechs page](http://ircv3.net/docs/sasl-mechs.html)

Issue #49, opened 2015-12-17
List information for services packages


DanielOaks:
As mentioned in #48, it would be nice to see information for services packages.

I'm not sure how much use services packages have for the various IRCv3 specifications (since they usually use an S2S protocol, probably a lot of things they don't care about), but at least for things like SASL support it would be very useful.

Mikaela:
:+1: as I started wondering about SCRAM-SHA-256 and if anyone supports it and such table could tell it to me immediately.

SaberUK:
> I started wondering about SCRAM-SHA-256 and if anyone supports it 

Nobody does currently but Anope will soon when I finish writing the module for it.

DanielOaks:
I'll consider this done for now.

PR #51, opened 2015-12-17
Add services support list

DanielOaks:
This just adds the services support list as discussed in #49. I think these are the main services packages around -- are there any others we should be listing as well?

I don't have it showing a page in the proper 'Software' section, as services packages don't make use of the sort of extensions IRCv3 brings out generally. If this changes or there is more to add along these lines, we can do so later.

### DIFF
--- a/_data/sw_services.yml
+++ b/_data/sw_services.yml
@@ -1,0 +1,25 @@
+- name: Services Packages
+  software:
+    - name: Anope
+      link: http://anope.org
+      support:
+        SASL:
+          - external
+          - plain
+    - name: Atheme Forks
+      link: http://atheme.net
+      support:
+        SASL:
+          - external
+          - plain
+      sublinks:
+        - name: ChatServices
+          link: https://bitbucket.org/chatlounge/chatservices
+        - name: EZtheme
+          link: https://github.com/ElectroCode/eztheme
+        - name: Shaltúre
+          link: https://github.com/shalture/shalture
+        - name: Xtheme
+          link: https://github.com/XthemeOrg/Xtheme
+        - name: Zohlai
+          link: https://github.com/zohlai/zohlai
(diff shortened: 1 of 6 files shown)


Mikaela:
I think you forgot [ChatServices] (https://bitbucket.org/chatlounge/chatservices). 

Are not-SASL or IRCv3 compatible clients counted? There is e.g.https://github.com/dkorunic/hybserv2 which doesn't seem very alive.

attilamolnar:
I think atheme forks should be listed together because there is little difference between them considering the overall codebase and no difference in sasl afaik. If one of them differs it should be listed seperately of course.

attilamolnar:
> There is e.g.https://github.com/dkorunic/hybserv2 which doesn't seem very alive.

From their README:

> Hybserv2 is incompatible with IRCD Hybrid 8.x as they have decided to be TS3 incompatible. There is some grunt work needed to make it fully TS6 and I personally don't have any more time and interest to code or IRC and related to IRC. For me, IRC has long died, [...]

So no, it should not be listed as it is incompatible with the current version of the server it targets and its author no longer supports it.

DanielOaks:
Alrighty, we now group Atheme forks together in the table. Should be fine to merge in that case?

attilamolnar:
Please squash the commits (rebase if necessary)

Mikaela:
:+1: if this PR is currently gathering :+1:s or is there some reason for not being merged already?

DanielOaks:
I'll assume this is good to merge, as everything requested has been done.

timofonic:
Why is this still open?

DanielOaks:
This isn't, this was merged at the end of last year.

Services aren't listed on the ordinary software support lists, ie [these](http://ircv3.net/software/servers.html), as our specs apply to the Client protocol and don't really affect them very much. This info is used and linked on things like the [SASL](http://ircv3.net/docs/sasl-mechs.html) pages though!

PR #53, opened 2016-01-16
Updated information about Rust IRC library.

aatxe:
Since this was added, I've updated the Rust IRC library quite a bit. It should support everything except TLS. So, I've updated the information accordingly.


DanielOaks:
Awesome, thanks for submitting the new info!

PR #54, opened 2016-01-17
Update information about Sopel

embolalia:
6.2.0 adds support for a bunch of things. More is planned; we're tracking it [here](https://github.com/sopel-irc/sopel/issues/971).


DanielOaks:
Awesome, thanks very much for the updates!

PR #55, opened 2016-01-22
sw_clients: cap-notify is in 1.4 (stable)

Mikaela:


DanielOaks:
Cool, thanks.

Just as a note, that build fail is unrelated, refers to some dodgy Python stuff rather than a failure with the changes here.

PR #35, opened 2015-11-09
software: Separate and highlight the required cap columns

dequis:
Preview: http://dump.dequis.org/LKeRZ.html

Current: http://ircv3.net/software/clients.html

Closes #18 

>I think some kind of formatting would make it more clear what clients are required to have in order to be IRCv3.1 compatible.
>
>Or group these columns in the left and separate from the rest by more think vertical line

I also made missing caps that are optional a lighter shade of red, since it's not wrong to not implement those. The current one gives off the appearance that everyone is doing everything wrong (the table is mostly red)

PR #57, opened 2016-01-25
Add Moon Moon IRC bot

ChickenNuggers:
Yes, I plan on adding more later. This is just what I could get done now.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -515,3 +515,25 @@
           - server-time
         SASL:
           - plain
+    - name: Moon Moon
+      # ref: https://github.com/ChickenNuggers/Moon-Moon/blob/master/plugins/data.moon
+      #      https://github.com/ChickenNuggers/Moon-Moon/blob/master/plugins/output.moon
+      #      https://github.com/ChickenNuggers/Moon-Moon/blob/master/plugins/base.moon
+      #      https://github.com/ChickenNuggers/Moon-Moon/blob/master/irc.moon
+      link: https://github.com/ChickenNuggers/Moon-Moon
+      support:
+        v3.1:
+          - cap
+          - multi-prefix
+          - account-notify
+          - away-notify
+          - extended-join
+        v3.2:
+          - cap
+          - account-tag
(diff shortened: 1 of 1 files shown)


SaberUK:
I think we only add stuff to the client list if it supports all of the base specifications. Do you have plans to add SASL-3.1 support?

ChickenNuggers:
The bot currently does not support it; I was unaware that it must support all base specifications (mIRC, for example, does not). Eventually, I plan in adding support for everything listed in v3.2.

Mikaela:
> I think we only add stuff to the client list if it supports all of the base specifications.

Not true. See https://github.com/ircv3/ircv3.github.io/issues/17 / https://github.com/ircv3/ircv3.github.io/pull/35.

jwheare:
I don't object to software that being on the list that doesn't implement all the specs, but from looking at the code it seems like many of these specs are only partially implemented.

I suggest you go back to the specs, and do more testing against real world servers. As well as the major networks that support IRCv3; ZNC, InspIRCd and mammon from git are good choices to test with for 3.2 CAPs. People in #ircv3 can point you to some test instances for these.

By no means intended a thorough code review, but a couple of concerns from looking at your source:

* account-tag only seems to be supported on the JOIN command, it should be parsed on every command originating from a user (easier just to parse it on all commands).
* cap-notify support appears to perform a CAP LS for every NEW CAP received (rather than checking if it's locally supported and doing a CAP REQ).

Happy to add you to the list in principle, even without all core specs implemented, but as it looks like you're still actively working on support and fixing things, so this PR seems premature.

Let us know when you've tested and bundled these changes into a tagged release so we can assign a version number to the features you support.

DanielOaks:
I'd agree with @jwheare here.

jwheare:
Closing for now. Feel free to reopen when this is no longer WIP and has a tagged release.

ChickenNuggers:
Release v0.2.0 is available and includes everything listed in the PR.

attilamolnar:
Please squash this before it is merged.

PR #58, opened 2016-01-25
travis: Use Trusty to fix Py builds

DanielOaks:
Don't actually merge this PR, I'm using this to test and see whether Travis builds properly with it. If it works, I'll just merge it directly.


DanielOaks:
Fixed, rebased, and merged into master.

PR #70, opened 2016-02-21
Add txircd to the list of compliant server software

Heufneutje:
Adding on behalf of @ElementalAlchemist.

We currently don't support SASL yet. The module for it has been written, however no mechanisms have been implemented yet.

### DIFF
--- a/_data/sw_servers.yml
+++ b/_data/sw_servers.yml
@@ -117,6 +117,28 @@
           tls: 2013-05-26
         v3.2:
           userhost-in-names: 2011-12-19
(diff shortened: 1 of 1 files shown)


DanielOaks:
This looks fine to me.

Thanks for the PR! Feel free to submit another one when the SASL mechanisms are implemented. You can also list the specific SASL mechanisms like [this](https://github.com/ircv3/ircv3.github.io/blob/master/_data/sw_services.yml#L5), which will get displayed on the website eventually.

PR #79, opened 2016-03-01
Change "IRCv3" in page title from prefix to suffix

jwheare:
This makes browser tab titles easier to compare when many tabs are open.


DanielOaks:
Sure, this makes sense.

PR #85, opened 2016-03-20
servers: Add BitlBee 3.4.2

dequis:


jwheare:
Possibly belongs in a gateways section, but fine here for now.

PR #87, opened 2016-03-27
Add BlueIRC

MicroDroid:


_data/sw_libraries.yml
```diff
@@ -80,3 +80,9 @@
           - userhost-in-names
           - batch
           - sasl
+    - name: BlueIRC
+      # ref: https://github.com/MicroDroid/blueirc/blob/development/src/tk/microdroid/blueirc/Handler.java#L59
+      v3.1:
```
DanielOaks: This should be under the `support:` key, similar to other libraries (see [this](https://github.com/MicroDroid/ircv3.github.io/blob/master/_data/sw_libraries.yml#L15) as an example). Otherwise, this looks good!

MicroDroid: oh I missed the support thing
I also forgot to add a link, I'll do the when  in my PC

_data/sw_libraries.yml
```diff
@@ -80,3 +80,11 @@
           - userhost-in-names
           - batch
           - sasl
+    - name: BlueIRC
+      # ref: https://github.com/MicroDroid/blueirc/blob/development/src/tk/microdroid/blueirc/Handler.java#L59
+	  link: https://github.com/MicroDroid/blueirc
+	  support
```
attilamolnar: Tabs here instead of spaces, please fix

attilamolnar:
This is still malformed, please fix.

attilamolnar:
Please rebase this commits so there's only one.

attilamolnar:
Thanks!

PR #88, opened 2016-04-13
clients: Update KVIrc

staticfox:
Added account-notify, away-notify, extended-join, chghost,
and userhost-in-names.


DanielOaks:
Thanks!

PR #90, opened 2016-04-17
_irc/3.1.md: Add RPL_ISUPPORT link

ChickenNuggers:
RPL_ISUPPORT is largely supported by IRCv3 members and afaik is required when supporting the IRCv3 protocol.

### DIFF
--- a/_irc/3.1.md
+++ b/_irc/3.1.md
@@ -2,7 +2,7 @@
 layout: page
 title: IRCv3.1
 ---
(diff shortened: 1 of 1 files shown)


DanielOaks:
The problem is that that document doesn't accurately represent what is implemented by software today. Especially because the `-<token>` syntax isn't implemented by clients and it lists a bunch of tokens that don't really get advertised it makes me hesitant to link to it.

I believe someone was looking to write up a more modern `RPL_ISUPPORT` spec for IRCv3 to use / link to rather than linking to that draft?

ChickenNuggers:
If someone is working on a more modern version of RPL_ISUPPORT, then I'll close this; however, I'll keep my branch open for merging anyways, just in case.

PR #91, opened 2016-04-19
Add Kitteh IRC Client Library to Libraries (and sort them alphabetically)

mbax:
Two commits for separation of tasks.


DanielOaks:
Thanks, and thanks for sorting them!

PR #93, opened 2016-05-27
Add the lounge web client entry

omnicons:
As requested by @YaManicKill and based on their [ircv3 compatibility](https://github.com/thelounge/lounge/issues/312)


DanielOaks:
Thanks!

PR #94, opened 2016-05-28
Add "The" in "The Lounge"

omnicons:
As per the request of one of the developers


astorije:
FYI, this is the official name. See https://thelounge.github.io/.

PR #95, opened 2016-05-28
Switch URL for The Lounge from GitHub repo to official website

astorije:
This is to be consistent to all other entries.
Also, this is the entry point for the software.


DanielOaks:
Thanks, this has been merged into master!

PR #99, opened 2016-06-17
Fix responsiveness of the main page

xPaw:
"IRCv3 Features" lists and below.


DanielOaks:
Sure, this works. Thanks!

PR #100, opened 2016-07-09
docs/SASL-mechs: Add Libraries

Zarthus:
Adds SASL mechanisms for libraries to the best of my abilities. The docs in most cases weren't particularily clear on EXTERNAL or other sasl mechs support. Some libraries like cinch support blowfish but weren't mentioned due to them being deprecated.

It might be worth considering adding another column for 'ecdsa-nist256p-challenge', feel free to comment thoughts about that.

Caps are alphabetically sorted.

sasl mechs

### DIFF
--- a/_data/sw_libraries.yml
+++ b/_data/sw_libraries.yml
@@ -8,6 +8,8 @@
           - cap
           - multi-prefix
           - sasl
+        SASL:
+          - plain
     - name: cinch
       # ref: https://github.com/cinchrb/cinch/blob/v2.3.1/lib/cinch/irc.rb#L115
       link: https://github.com/cinchrb/cinch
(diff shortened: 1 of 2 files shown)


_data/sw_libraries.yml
```diff
@@ -60,6 +66,10 @@
           - sasl
           - server-time
           - userhost-in-names
+        SASL:
+          - ecdsa-nist256p-challenge
```
SaberUK: This is not on the table so there's no need to add it here.

Zarthus: Commit amended, based on the discussion on IRC it seems unlikely it is going to be added.

DanielOaks:
Neat, thanks for the PR!

PR #104, opened 2016-10-30
data/sw_*: change name `tls` to `starttls`

ChickenNuggers:
As discussed on IRC, using the name `tls` gave the impression this meant the TLS protocol rather than the TLS specification and STARTTLS command. This commit should change the header to `starttls` rather than `tls`.


DanielOaks:
Looks good, thanks very much

PR #106, opened 2016-11-07
Networks support page

DanielOaks:
Very basic for now, but we'll extend it out more later. Opinions before I merge this in?

I'll also fix the build failure.

networks - ircv3-2

### DIFF
--- a/_data/irc_versions.yml
+++ b/_data/irc_versions.yml
@@ -92,3 +92,21 @@ v3.2:
       name: userhost-in-names
       description: userhost-in-names Extension
       link: http://ircv3.net/specs/extensions/userhost-in-names-3.2.html
+
+v3.3:
+  name: IRCv3.3 Draft
+  note: >
+    **These specifications are drafts and are NOT completed IRCv3 specs.**
+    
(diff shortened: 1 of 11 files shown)


dequis:
The networks section needs a note clarifying that it's an incomplete list of networks whose opers agreed to be mentioned in that page and that it doesn't include all the networks supporting ircv3 features.

It's tempting to throw a number there to give users an idea of how widespread the whole thing is - your stats page says 70% which, lacking weighting by user count, is a conservative number IMO. But it's probably better to say it in a way that doesn't imply confidence, something like "most networks".

Zarthus:
Something I just brought up on IRC; but I'd personally like to see drafts and experimental goodies listed here as well.

````
(22:42:36) <Zarthus> is sts support not a capability/worth mentioning on the website (at least for testnets; specifically directed to #106)
(22:43:24) <@jwheare> It's still a draft.
(22:44:00) <@jwheare> Maybe drafts can be listed for testnets only.
(22:44:07) <Zarthus> might still be worth mentioning in the testnets
(22:44:22) <Zarthus> I think inspircd at least implements STS, which is worthwhile for testing implementations
(22:49:49) <@jwheare> And message tags
```

PR #108, opened 2016-11-14
Add zIRC to the libraries list

wolfy1339:


_data/sw_libraries.yml
```diff
@@ -138,3 +138,16 @@
...
+      # ref: https://github.com/itslukej/zirc/blob/master/zirc/caps.py
+      #      https://github.com/itslukej/zirc/blob/master/zirc/ext/sasl.py
+      #      https://github.com/itslukej/zirc/issues/9
+      link: https://github.com/itslukej/zirc
+      support:
+        v3.1:
+          - cap
+          - sasl
+        v3.2:
+        - message-tags
```
DanielOaks: This line should also be indented like the 3.1 lines are

wolfy1339: I put it there because another library had it listed at the same spot

attilamolnar: Right, that's an oversight. Will be fixed shortly.

DanielOaks:
Thanks very much!

PR #109, opened 2016-11-20
Add basic FAQ

DanielOaks:
People have mentioned that this is a good thing to have, and I agree. Answers some basic q's about us, etc.

FAQs


jwheare:
Nice. Good start. I think we shouldn't assume the questions will be quite so inflammatory though :) We can address the same points with more neutral language I think.

DanielOaks:
Yeah @dequis suggested that as well, so far I've got:

s/Why are you ignoring older clients?/What about older clients?/
s/Why haven't you made any specs for the server-to-server protocol?/What about the server-to-server protocol?/

I'll think about it as well but any other suggestions, feel free to throw them in here or at me in the channel

jwheare:
s/Why are you destroying IRC/Is IRCv3 backwards compatible/

jwheare:
Basically, instead of accusatory why questions, just questions that have a non-defensive answer that can be expanded on.

Who decided you get to control -> Who runs IRCv3
Why doesn't RFC make sense -> Are you working on an RFC?
Why haven't you done s2s -> Are you working on the s2s protocol too?
When will feature be done -> Is there a roadmap or timeline? (link to Roadmap milestone)

DanielOaks:
Alright this should be good to go now, just removed the last q.

FAQ

jwheare:
LGTM

PR #110, opened 2016-11-20
su_networks.yml: lowercase freenode's name

Xenthys:
A tiny PR, but some people actually care about freenode's case.


DanielOaks:
Makes sense, thanks for the PR!

PR #111, opened 2016-11-20
supported networks freenode: change irc.* to chat.*

Zarthus:
I'm not *confident* if that's changed recently, but as far as I know they prefer chat.*


DanielOaks:
Yeah they have that as the main addy on their site too, makes sense

PR #114, opened 2016-11-20
su_networks.yml: Add IRCHighWay

dequis:
I asked in the super secret irchighway dev channel a while ago:

```
21:38 < dx> hey do you want to be added here http://ircv3.net/support/networks.html
21:50 <@polyn> dont see why not
22:06 < Andrio> I vote for.
```

(They are netadmin and globalop respectively)


DanielOaks:
Sounds good, thanks very much for checking with them

Issue #103, opened 2016-10-10
Occasional news releases with new specs/proposals/etc


DanielOaks:
This was brought up again recently, just like a news tab on the site with a simple RSS feed to announce changes to interested people. Could be cool to look at doing, would probably throw it on the WG page to improve #83 as well.

Zarthus:
a status page on the front page with a list of specs and the current status would be nice as well

DanielOaks:
Solved by #115

Issue #83, opened 2016-03-11
Improve Working Group page


DanielOaks:
I think there's quite a bit we could do to extend this page and make it more useful for people who are a part of or explicitly interested in joining the working group.

Improving the [participation page](http://ircv3.net/participation.html) to include the technical board details in the [charter](http://ircv3.net/charter.html) would probably be a part of this (adapting it to use a `_data` list instead of being defined just on that page so we can add it there too, as well as just improving the look of the thing).

As well as that ircv3 announcements mailing list idea that was suggested, the roadmap stuff that's being done in https://github.com/ircv3/ircv3-specifications/issues/241 and other details we could add or link to this WG page for interested people.

DanielOaks:
Solved by #115

PR #115, opened 2016-11-20
Add RSS feed with basic spec updates

DanielOaks:
For some time, people have wanted a way to just subscribe for the occasional update on what we're doing. This PR introduces an RSS feed to let people subscribe for the occasional update on what we're doing, the status of our specs, and recent changes.

This also fills out the WG page a fair bit, which is good. Solves #83 and #103.

Example post:
screen shot 2016-11-20 at 12 35 41 pm

And how it's displayed on the WG page:
screen shot 2016-11-20 at 12 35 27 pm

### DIFF
--- a/wg.html
+++ b/wg.html
@@ -33,3 +33,25 @@ meta-description: Welcome to the IRCv3 Working Group. We're a group of IRC clien
           <a class="btn btn-lg btn-primary" href="https://github.com/ircv3">View</a>
         </div>
       </div>
+
+</div>
+
+<div id="home-slider-wrapper"><div class="container">
+
+      <h2>IRCv3 WG Updates <a style="padding: 0.2rem 0.6rem; border-radius: 0.4rem; background: #F2F2F2; margin-left: 1rem; font-size: 1.85rem; color: #cc6f0f" href="/feed.xml"><i class="fa fa-rss-square"></i> RSS</a></h2>
+
+      <p style="margin-bottom: 2rem">Below is a rough log of what the IRCv3 working group's been up to including recent changes to specifications and new proposals:</p>
+
+</div></div>
+
+<div class="container">
+
+      {% for post in site.posts limit:10 %}
+      <div class="post">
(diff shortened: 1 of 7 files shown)


jwheare:
Great! Let's get a link to the roadmap milestone on the WG page while we're at it. That and status updates/changelog should probably be on the home page too.

DanielOaks:
Yeah threw roadmap link on the WG as 'the top article', should go well. I'm iffy on throwing it on the home page too, just due to how much we already have on there and having to modify this a decent amount to get it to play nicely there. Mind if we leave that 'til after this is merged/in, since it may take a bit of time to play with and see exactly how we wanna tackle that?

wg.html
```diff
@@ -33,3 +33,32 @@
...
+      <p style="margin-bottom: 2rem">Below is a rough log of what the IRCv3 working group's been up to including recent changes to specifications and new proposals:</p>
+
+</div></div>
+
+<div class="container">
+
+      <div class="post">
+          <h3><a href="https://github.com/ircv3/ircv3-specifications/milestone/4">Roadmap</a></h3>
+          <div class="post-content">
+              <p>Our roadmap <a href="https://github.com/ircv3/ircv3-specifications/milestone/4"><sup>[link]</sup></a> always contains info on our direction moving forward, the sort of features we have planned and our priorities.</p>
```
dequis: `roadmap <a><sup>[link]</sup></a>` -> `<a>roadmap</a>`?

jwheare:
Yeah fine by me to delay home page stuff.

DanielOaks:
Cool, cleaned up the language there and added notes for STS and SNI.

screen shot 2016-11-20 at 2 00 37 pm

Should be good to go with this?

ChickenNuggers:
Is there a reason links are not implemented using just anchor tags for `[link]`s?

DanielOaks:
Because they're a bit difficult to see and this brings them out a bit more. I'll make links in general easier to see but I'd probably prefer having the external links like this (particularly as this is kinda like a reference page, having the explicit external links could make sense here).

ChickenNuggers:
> having the explicit external links could make sense here

Now that you mention it, that's a good point.

jwheare:
LGTM

PR #117, opened 2016-11-20
_data/su_networks.yml: change formatting from null-dicts to array

ChickenNuggers:
Testing Website

Basically, the formatter reads the **values** of an array, instead of the **keys** of a dict containing nil/null values.


DanielOaks:
Makes sense for the networks page, since they're not gonna be listing dates or etc. Cool, thanks for the PR!

PR #119, opened 2016-11-20
_irc: squash documents and add description

### LINKED ISSUE #82: Improve Specs page
I feel like we could make [this page](http://ircv3.net/irc/) quite a bit better.

Especially with the discussions we've been having about the merits of actually releasing features in batches (3.1, 3.2, 3.3, etc), it may be worth looking at different ways we could display the features.

One thing we can look at is the [IRCCloud page](https://blog.irccloud.com/ircv3/) describing the various IRCv3 features, and whether it'd be useful or appropriate for us to take a similar approach (presenting each spec with a short tagline describing what it accomplishes and giving a general overview of it). Depends on our audience and who we're targeting with the page, but I'll sketch out some ideas, and people can present their own in here.

ChickenNuggers:
This is an extremely rough draft meant to help fix #82 and #9 - please suggest amendments.

### DIFF
--- a/_data/su_networks.yml
+++ b/_data/su_networks.yml
@@ -45,6 +45,7 @@
           - starttls
         v3.2:
           - account-tag
+          - batch
           - cap-notify
           - cap
           - chghost


ChickenNuggers:
btw you can view the changes [here](http://ny1.hashbang.sh:25568/irc/).

DanielOaks:
Bloody Git doing weird things when branches are named differently. I've pushed my changes to the language, intended to put forth the case for extensions more strongly, link related specs more closely and remove unnecessary differences like the diff between core and extension specs... at least, the _visible_ differences. It's clear that one section is more important than the other which is why we keep them separated in our lists and in this page, but outlining those differences isn't as useful as just explaining the specs themselves, I think.

New PR over here: #120

PR #120, opened 2016-11-20
Overhaul specifications page

DanielOaks:
So we're overhauling the specs page to better represent how we want to display specs to users. This gives people more of a reason to implement and care about them, and plays into separating specifications from the 'version' cycle as discussed in https://github.com/ircv3/ircv3-specifications/issues/265

Thanks goes to @ChickenNuggers for starting the work on this, much appreciated.

new specs page


_irc/index.md
```diff
@@ -5,12 +5,276 @@ excerpt: IRCv3 Specifications
...
+
+The [v3.2 Capability Negotiation spec](http://ircv3.net/specs/core/capability-negotiation-3.2.html)
+clarifies how to list large numbers of capabilities (which is needed as we add
+more useful features to IRCv3!). It also allows capabilities to be listed with
+values, improving the implementation of features like SASL.
+
+The [`cap-notify` spec](http://ircv3.net/specs/extensions/cap-notify-3.2.html)
+allows clients to be sent notifications when caps are added to or removed from
+the server. This is useful in cases like SASL when the authentication layer
+disconnects (and thus, SASL authentication is no longer possible). As a note,
```
ChickenNuggers: Technically it's not.

DanielOaks:
Cool, made it more alphabetical (while listing caps and message-tags right up the top) and removed that "as a note" text.

New Specs Page

DanielOaks:
As a note, the CI failure is unrelated and already fixed in master by 7754b44206112232288b21142833f1a2d3ce1ee5

lol768:
Looks much more organised in the screenshot and I prefer the overall approach to the current one.

Does it really need to say the spec is a draft _three_ times in bold all-caps, though? It distracts from the actual description of what the spec does and just seems excessive. The draft specs could be put in their own section or prefixed with [draft] or something.

ChickenNuggers:
:+1: for [draft]

DanielOaks:
I can try some other ways to note draft specs, definitely. The main thing is that I want it to be very explicitly clear, which is why it goes overboard right now. But we can make it very clear in ways like having a small note when you mouse over it and/or just experimenting with the display more, so I'll give those a shot and let you know. Thanks for the suggestions, much appreciated.

ChickenNuggers:
While we're doing this, can we replace the static URLs with dynamic ones? `sed -i 's|http://ircv3.net||' _irc/index.md` should work as a simple replacement script.

DanielOaks:
Coolcool, using better links and the **work-in-progress** language rather than the mess of all-caps junk that was there before. WIP is clearer language than draft, and we only need to say it once.

IRCv3 Specs

_irc/index.md
```diff
@@ -5,12 +5,273 @@ excerpt: IRCv3 Specifications
...
+
+
+## [Message Tags]({{site.baseurl}}/specs/core/message-tags-3.2.html)
+
+Message tags extend the core framing used with IRC messages, and allow extended
+data to be send with messages. Tags are used by a number of extensions, and we
+consider them a core part of IRCv3 and of supporting new extensions.
+
+This feature is considered essential and all software implementing IRCv3
+**must** implement at least the core Message Tags specification.
```
jwheare: Don't think we should say this. The whole point of dropping the core/extension and version distinctions is that you should be able to implement IRCv3 as you choose. I don't think implementations that don't support any of the tag specs should be considered non-compliant.

DanielOaks:
@jwheare both of those make sense, added the changes

DanielOaks:
Alright cool, I'll throw this in in the next few days if there's no further comments and we're happy to do so.

IRCv3 Specs
.

DanielOaks:
Merged into master!

PR #122, opened 2016-12-09
Update zIRC supported CAP

wolfy1339:


DanielOaks:
Nice, thanks very much for the PR!

PR #123, opened 2016-12-11
Add ChatZilla to client table

twpol:
We only support multi-prefix right now but are planning to expand. :)


DanielOaks:
Nice, thanks very much for the PR! :)

PR #124, opened 2016-12-17
chghost is supported in AdiIRC

tbs:
https://dev.adiirc.com/projects/adiirc/wiki/IRCv3


DanielOaks:
Thanks for the PR, much appreciated!

PR #125, opened 2016-12-17
sw_services: rename "Atheme forks" to "Atheme (and forks)"

GLolol:
Atheme is in active development again, so not including it is a bit misleading.


DanielOaks:
Makes sense, thanks for the PR!

PR #126, opened 2016-12-21
irc: Rewrite header, link to Modern docs

DanielOaks:
The current _"IRCv3 Specifications"_ intro is old and a bit crufty. This PR rewrites it to flow a bit more nicely.

In addition, as suggested in https://github.com/ircv3/ircv3-ideas/issues/3 it links to the [Modern docs](http://modern.ircdocs.horse/) as a possible resource to look at. The things to keep in mind here are that the Modern docs are explicitly edited by me, which is why I've tried to use language distancing it from being an IRCv3-created thing.

Here's how it looks:

IRCv3 Specifications

### DIFF
--- a/_irc/index.md
+++ b/_irc/index.md
@@ -8,16 +8,19 @@ index: true
 
 IRCv3 specifications build on top of the core IRC protocol. The primary core
 IRC protocol specifications are [RFC1459](http://tools.ietf.org/html/rfc1459),
-[RFC2812](http://tools.ietf.org/html/rfc2812) and [RFC7194](https://tools.ietf.org/html/rfc7194),
-although one of our members has been working on updated core protocol
-specifications [here](http://modern.ircdocs.horse/) which you may find useful
-to consult.
+[RFC2812](http://tools.ietf.org/html/rfc2812) and [RFC7194](https://tools.ietf.org/html/rfc7194).
+To fully understand IRCv3, please read the core IRC protocol specifications
+followed by the IRCv3 specifications.
+
+**Note:** The core IRC protocol specifications have been widely acknowledged to
+be old, out-of-date, and to not fully or accurately describe how the IRC client
+protocol works today. One of our members has been working on updated core
+protocol specifications [here](http://modern.ircdocs.horse/) which you may find
+useful to consult. Look at core protocol specifications and the IRC ecosystem
+itself for a complete picture of how the IRC protocol works today.
 If you have any questions on the core IRC protocol, please feel free to ask us
 [directly]({{site.baseurl}}/contact.html) or with an issue in our
 [feedback](https://github.com/ircv3/ircv3-feedback) Github repository.
-
-To fully understand IRCv3, please read the core IRC protocol specifications
-followed by the IRCv3 specifications.
 
 These specifications are released when they are stable and have been widely
 tested. In the past, the WG released specifications as versioned bundles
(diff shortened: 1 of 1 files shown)


_irc/index.md
```diff
@@ -6,15 +6,21 @@ meta-description: Specifications that the IRCv3 Working Group maintains and dist
...
-[RFC1459](http://tools.ietf.org/html/rfc1459). To understand the basis of the
-IRC version 3 protocol, please read RFC1459 followed by the extension
-specifications.
-
-IRCv3 is based on a set of specifications. Initially these specifications were
-released as a versioned distribution – these days, however, each specification
-is released when it is stable and has been widely tested.
+IRCv3 specifications build on top of the core IRC protocol. The primary core
+IRC protocol specifications are [RFC1459](http://tools.ietf.org/html/rfc1459)
+and [RFC7192](https://tools.ietf.org/html/rfc7194), although one of our members
```
jwheare: 7192 -> 7194 in the link text.

jwheare:
Should link to 2812 too.

_irc/index.md
```diff
@@ -6,15 +6,21 @@ meta-description: Specifications that the IRCv3 Working Group maintains and dist
...
+has been working on updated core protocol specifications
+[here](http://modern.ircdocs.horse/) which you may find useful to consult.
+If you have any questions on the core IRC protocol, please feel free to ask us
+[directly]({{site.baseurl}}/contact.html) or with an issue in our
+[feedback](https://github.com/ircv3/ircv3-feedback) Github repository.
+
+To fully understand IRCv3, please read the core IRC protocol specifications
+followed by the IRCv3 specifications.
+
+Initially these specifications were released as a versioned distribution, but
```
Zarthus: That sentence is going to date quickly, maybe something like "up until version 3.2 (or a date?), specifications were released as a versioned distribution. This is no longer the case and each specification is now released when ..."

or "The specifications are released when they are stable and have been wildly testing. In the past, IRCv3 adopted a versioning scheme..."

DanielOaks: Makes sense, changed it around to push what we do right now instead

JonasT:
I find the suggestion a good idea regarding the basic motivation, but to me the suggested wording seems pretty ambiguous to what actually makes the specs, what should be read and also *why* it should be read.

Therefore I would suggest first stating clearly what the current formal specifications actually are, and then afterwards, as a clear separate paragraph, explain why the updated draft exists and why it should be read (-> a clear remark that the RFCs are outdated).

> IRCv3 specifications build on top of the core IRC protocol. The primary core IRC protocol specifications are RFC1459, RFC2812 and RFC7194.
>
> To fully understand IRCv3, please read the core IRC protocol specifications followed by the IRCv3 specifications.
>
> Note: we are aware that the primary core IRC protocol specifications are outdated and do not always fully apply in modern IRC anymore. One of our members has been working on updated core protocol specifications here which you may find useful to consult. If you have any questions on the core IRC protocol, please feel free to ask us directly or with an issue in our feedback Github repository.

DanielOaks:
I can see splitting it out a bit more nicely. I think this reads a bit better and conveys things more accurately

JonasT:
Neat! Minor issue/nitpick I'm still seeing:

```
These specifications are released when they are stable and [...]
```
.. follows up the note paragraph about the core specs but actually talks about the IRCv3 specs, so for clarity it should probably be
```
The IRCv3 specifications are released when they are stable and [...]
```

DanielOaks:
Cool, looks fine. We'll adjust it later as necessary if we need to

PR #128, opened 2017-01-05
libraries: KICL now supports draft/sts and draft/sni

Zarthus:


attilamolnar:
These are drafts and shouldn't be on the support pages

PR #130, opened 2017-01-05
IRCv3.3 drafts: add SNI

Zarthus:
Can anyone verify whether the `required` field is set appropriately?


attilamolnar:
SNI is required and there are no caps associated with it

DanielOaks:
Thanks for the PR but I'll hold off on this 'til bigger support page changes. SNI's a bit of a weird case anyways (more like it fits in as a 'docs' thing rather than a proper spec, to me at least).

PR #132, opened 2017-01-06
services: Remove EZtheme

attilamolnar:
EZtheme hasn't been updated since 2015-02-12, shortly after the https://github.com/ElectroCode/eztheme repo was created on 2014-12-29

    <SaberUK> it might be a good idea to clean out the list of atheme forks 
    <SaberUK> because some of them are dead
    <SaberUK> eztheme hasnt had commits in two years, shalture/zohlai hasnt had commits in a year,

### DIFF
--- a/_data/sw_services.yml
+++ b/_data/sw_services.yml
@@ -15,8 +15,6 @@
       sublinks:
         - name: ChatServices
           link: https://bitbucket.org/chatlounge/chatservices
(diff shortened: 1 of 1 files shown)


attilamolnar:
Source of dates: https://api.github.com/repos/ElectroCode/eztheme

DanielOaks:
Makes sense, thanks

PR #135, opened 2017-01-13
New Support Lists

DanielOaks:
Recently, we unbundled specs from explicit 3.1/3.2/etc releases on the specs page. However, our support lists still show everything as being under 3.1, 3.2, etc.

These new support lists are more extensible and combine the specs in a nice, easy-to-understand way that works better for how we're presenting and releasing specifications these days.

### Examples

Clients page:
IRCv3 Clients page

Servers page:
IRCv3 Servers page

---------

Thoughts?


DarthGandalf:
Generally, looks good!
Let's not hide the all-no rows. E.g. I don't see the roadmap what ZNC doesn't implement yet.
Ideally, the table should be split dynamically, depending on window size. There's lots of free space to both sides of screen.

DanielOaks:
Yeah dynamically splitting it requires javascript and all that fun stuff. If someone wants to take a shot, they're free to go ahead, but ye.

I agree with the all-no rows -- I think that was just a side-effect of my fun with the support list file. Will fix

DanielOaks:
Actually, I've made it the original width again. Here are some new screenies, looks a decent bit better:

Client page:
IRCv3 Client page

Server page:
IRCv3 Server page

Zarthus:
I find the list a bit overwhelming and especially for the right-most entries in desktop and mobile it gets hard to keep track. I assume that when you hover over it with the mouse it makes it more visible but navigating the list via the screenshot made my eyes hurt a little bit.

Maybe splitting up the lists in multiple counterparts if they exceed 6 entries would help? I'm not sure. I just think it shows too much compacted information right now.

DanielOaks:
Hmm, this is an example of what it looks like when you hover over an item:

new-support-list-hover

Solving that is difficult, because we need to convey a whole lot of info at once. I'll have a play, maybe manually marking tables to be 100% when they exceed a certain point or something similar.

Zarthus:
(probably for a different issue)
I also think now would be a good time to ditch some of the inconsistent numbers like with ZNC having 0.094 ...

I would go with "yes" for things that aren't in git, and assume the user always uses the latest stable software or oldstable. It's not ircv3's concern to keep a changelog of each client.

DanielOaks:
Made the cols wider because it was difficult to read them (thanks @Zarthus for pointing it out!), so should be a fair bit nicer now. In addition, it now takes up the full width if it should (if the table has like 8 s/w columns or more).

Client page:
IRCv3 Clients page

DanielOaks:
Cool, I'll merge this tomorrow if there aren't any further issues with it

jwheare:
Can you do anything about the text shifting by a pixel when you hover? I think the tooltip whitespace is interfering.

lrstanley:
if this does get implemented, may be worth while to implement [pull/139](https://github.com/ircv3/ircv3.github.io/pull/139) into this as well, which adds support for showing server/library languages being used.

DanielOaks:
This isn't being merged right now. @attilamolnar has brought up concerns with the way that this displays specs wrt the unbundling and all. I'm giving peeps time to put forth some alternate suggestions on this page and the specs page (since the display here matches the specs page, if there's something wrong here we need to fix it on the specs page first).

jwheare:
Cheers Dan, shouldn't be hard to sort out. I'll have some time to chat after Monday.

dequis:
FWIW I just noticed that the current version in master doesn't have any caps for chatzilla, but this PR fixes it in the sw_clients.yml part of b005cfd25b268b28ad95d54155b966dd412d8fb7. May be worth backporting that to master if this PR is on hold.

DanielOaks:
I'll resolve conflicts after things are worked out here

DanielOaks:
Aight so this should be fine to merge now. If there aren't any further concerns I'll merge this in the next few days.

jwheare:
Pushed some commits to fix labeled-response, updated the message-tags-0.2 version, and updated the IRCCloud level of support.

Live preview version here:
http://jwheare.github.io/ircv3.github.io/support/networks.html
http://jwheare.github.io/ircv3.github.io/software/clients.html
http://jwheare.github.io/ircv3.github.io/software/servers.html
http://jwheare.github.io/ircv3.github.io/software/libraries.html

PR #138, opened 2017-01-16
Use Bundler for gem version management

CarrotCodes:
Benefits include being able to match GitHub Pages dependency versions for local testing, and slightly easier setup (assuming you have Bundler, which is reasonable).

Also updates README with Bundler instructions.

When you update the `Gemfile`, you should `bundle install` and commit the `Gemfile.lock` too.


DarthGandalf:
Someone will need to make sure Gemfile.lock is up to date.

jwheare:
I'm happy to keep it up to date, it shouldn't require frequent changes.

DanielOaks:
I don't use Bundler and all, but if someone who does confirms this looks good and works fine then feel free to merge

PR #139, opened 2017-01-18
add programming language info to servers/library lists

lrstanley:
basically looks like (for both libraries, and servers -- potentially more if one wanted?):

different colors for the different languages.

also of note, this uses titles so hovering over would provide the language as well.

### DIFF
--- a/_data/sw_libraries.yml
+++ b/_data/sw_libraries.yml
@@ -1,6 +1,7 @@
 - name: Libraries
   software:
     - name: cinch
+      language: Ruby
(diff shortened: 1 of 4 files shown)


lrstanley:
basically looks like (for both libraries, and servers -- potentially more if one wanted?):


different colors for the different languages.

DarthGandalf:
Name of language itself would be easier than trying to look and match colors in legend all the time. Especially for color-blind.

prawnsalad:
Note that irc-framework runs both on NodeJS and the browser.

PR #142, opened 2017-01-28
Add ecdsa-nist256p-challenge to the list of SASL mechanisms

ProgVal:


Zarthus:
I think they didn't want to add this because it wasn't specified somewhere. I'm not sure of the exact reason, but it's omission is intentional (or I'm confused with a different mechanism)

Zarthus:
https://github.com/ircv3/ircv3.github.io/pull/11

PR #144, opened 2017-01-31
Update Warren links to WillowChat organisation, add newly supported caps

CarrotCodes:
As title describes. Not sure about the support list changes, but I'm happy to do it again after if it becomes unblocked.


jwheare:
Thanks! There are a few other support list PRs recently that @DanielOaks is already keeping track of, we'll make sure this gets included.

PR #152, opened 2017-03-17
New look and font

DanielOaks:
So a little while back I wanted to have a new logo for IRCv3, something proper to replace the current beloved [identicon](https://github.com/blog/1586-identicons). So jwheare and I got to work making up a new logo and giving us a properly interesting look and feel.

Here's the new logo we've workshopped up, in the various bg/fg colour combinations it can come in:

sidebyside

Our current website looks like a stock Bootstrap/etc themed site. The fonts used on it evoke more of a "this is a stock website" feel than anything. I think this PR gives us a fresher look with a bit more character.

I've got an online version browsable here: http://danieloaks.net/ircv3-site/

Thoughts on this? Which parts of the new look are an improvement, and which parts aren't as good? How's the readability, how do you think it feels overall?

Interested in feedback on this.


Zarthus:
I think the third one should not be used (at least in comparison to the other two); it has a different "background" colour than the other two, inconsistent with the theme.

prawnsalad:
The V doesnt look like a V if you didnt know of IRCv3 already. It just looks like decoration to me

CarrotCodes:
I like the font, and the logo (the 'v' is fine looking like decoration, the logo should convey "IRC version 3" and "IRC 3" does that fine regardless). I'm probably biased because the right hand side sort of looks like a :heart:.

The sizing and kerning of fonts across the home page make the flow very odd for me. The title font seems very horizontally squashed and could use some extra space between letters. The title appears very big, as do the three sub-headers ("Working Group" etc), but their descriptions look small in comparison, and the amount of padding on the "View" buttons doesn't help either.

I think it's probably worth considering the hierarchy of information you're trying to convey and either tweaking font sizes, or reorganising sections, to make it flow properly.

DanielOaks:
That's fair, I'm considering this more of an initial step to introduce the new logo and typefaces. I'm playing with some concepts for a more extensive overhaul of the design that also reorganises the sections a bit and fixes up the information flow proper.

However, I'll take the font sizes in mind and see if I can make it all flow a bit nicer in this initial PR. Thanks for the feedback and for the specific examples!

DanielOaks:
Hmm, I'll actually just push that new design instead of this interim PR and then that separately.

PR #153, opened 2017-03-20
Design overhaul

DanielOaks:
So we've had the current site design for a fair number of years. A long, long time. Still based on the original look we inherited from Atheme, and very much a stock-ish Bootstrap site. The colours look a bit old-fashioned and the font choices aren't that inspired.

With this PR, I'm giving us a fresh, modern look, including a new logo and colour scheme workshopped by jw and I. This PR also brings a real separation between the landing and WG pages. I'll go through the changes separately.

Online version: http://danieloaks.net/ircv3-s2/

### [Landing Page](http://danieloaks.net/ircv3-s2/)

With the new design, this page properly represents somewhere that someone new to IRC can go. The IRCv3 homepage is commonly linked as something like a response to the question _"Isn't IRC just old and something nobody cares about? Why should I use it over Slack/etc?"_. This header gives a proper, short, sweet explanation of why we think IRC is good, and should be used today. The new Introduction section gives new developers some direction and advice on where to go if they're getting into it for the first time.

We also link to the FAQ here now, and even answer some of them in that introduction section, which is nice.

### [Working Group](http://danieloaks.net/ircv3-s2/wg.html)

This WG page is cleaned-up and focused. It links to our charter, participation page and our Github, which are the main WG assets people will want to checkout. The header better explains what we are (the old one just said _"Welcome to the WG"_, and pretty much said nothing), and the short Introduction section gives people a more in-depth explanation of what we are and what we do.

---------

In short, I think the changes made to this design much better explain what we are, give a clearer overview to new people joining us or just visiting the site, and just makes IRC (and IRCv3 in particular) look more favourable vs some of the other chat systems out there.


CarrotCodes:
I think this design looks great so far ✨. The information flows much better now the font sizes are more coherent.

I have two issues:
* The text seems 'one size' too big everywhere. A zoom of 90% (using Cmd Minus) makes it much more like what I'm expecting compared to other sites with lots of text (like Medium, for example)
* The line height in paragraphs is too small - it's hard to read. I experimented and a line-height of 1.4 seems perfect for my tastes, but maybe worth experimenting yourself.

I've attached an image of what those changes look like for me.
screen shot 2017-03-20 at 19 46 52

jwheare:
Any remaining blockers?

DanielOaks:
I'm prolly gonna size the font down just a touch overall and then merge it in, other than that no blockers. I can do that font change this weekend (tomorrow).

DanielOaks:
Aight I'm gonna merge this into master

PR #155, opened 2017-04-16
Updated support for IRC v3 in Konversation 1.7

petersimonsson:
Added supported features in Konversation 1.7


Zarthus:
While not entirely related to this PR, I've noticed the URL could also be replaced with a HTTPS one. Their cert seems to expire in 2019.

DanielOaks:
Nice, thanks very much!

PR #157, opened 2017-04-20
Add Mutter to supported mobile clients

MuffinMedic:


_data/sw_clients.yml
```diff
@@ -441,6 +441,22 @@
...
+      support:
+        stable:
+          cap-3.1:
+          away-notify:
+          batch:
+          multi-prefix:
+          sasl-3.1:
+          server-time:
+          userhost-in-names:
+      SASL:
```
DanielOaks: SASL should be indented two spaces in so it's under support, other than that looks great!

DanielOaks:
Thanks very much!

PR #161, opened 2017-05-26
Update kiwiirc client support

prawnsalad:


DanielOaks:
Nice, thanks very much!

PR #162, opened 2017-05-31
Update Mutter caps

MuffinMedic:


DanielOaks:
Nice, thanks very much for the PR!

PR #164, opened 2017-06-04
add girc to the list of supporting libraries

lrstanley:
https://github.com/lrstanley/girc/blob/master/cap.go
https://github.com/lrstanley/girc/blob/master/builtin.go

more support to come (e.g. sasl external, etc)


DanielOaks:
Thanks!

PR #166, opened 2017-08-15
Update support lists for IRCCloud, add client tags, message-ids -> msgid

jwheare:
* IRCCloud has monitor support and sts on the server now.
* msgid doesn't have a cap, but a tag, draft/message-ids doesn't exist
* removed plain undrafted `message-tags` from support lists
* add client tags to versions file and client support
* remove outdated note for echo-message about upcoming id specs that are now published (its spec needs updating to point to published labeled-response draft)

### DIFF
--- a/_data/irc_versions.yml
+++ b/_data/irc_versions.yml
@@ -118,13 +118,6 @@ stable:
       draft: true
       caps:
         - draft/labeled-response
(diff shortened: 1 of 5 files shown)


jwheare:
Er whoops, I pushed this by mistake. Comments still welcome :)

PR #167, opened 2017-08-26
Add ChatSharp to libraries list

RockyTV:
Depends on https://github.com/SirCmpwn/ChatSharp/pull/82 (fully support IRCv3.2 cap negotiation)


DanielOaks:
Nice, thanks for the PR!

PR #168, opened 2017-09-16
Fix chathistory batch link

mickael9:


DanielOaks:
Awesome, thanks very much for the fix!

PR #169, opened 2017-09-20
The Lounge supports more caps

xPaw:
Most of it comes by default from the IRC library we use.


DanielOaks:
Awesome, thanks very much for the PR!

PR #170, opened 2017-10-01
Update sw clients

Zarthus:
I went through most links where a reference was legible and tried to update the list. Most is already  up to date, though.

Visual representation: https://zarthus.github.io/ircv3.github.io/software/clients.html


DanielOaks:
Thanks, much appreciated!

PR #174, opened 2017-11-07
Palaver 2.7.4 supports sts

kylef:


DanielOaks:
Nice, thanks very much!

PR #175, opened 2017-11-08
Add Numerics section to registry page

DanielOaks:
Adds a section to the Registry page called Numerics, that simply lists numerics that we define or recommend using (such as `005` as `RPL_ISUPPORT`).

Right now, the ISUPPORT numeric links to the relevant Modern docs sections for it, but whatever works.

screen shot 2017-11-09 at 2 44 23 am

### DIFF
--- a/_data/registry.yml
+++ b/_data/registry.yml
@@ -1,4 +1,108 @@
+- name: Numerics
+  include_numeric: true
+  include_specs: true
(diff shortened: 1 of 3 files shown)


jwheare:
LGTM

DanielOaks:
Merged into master

PR #178, opened 2017-11-12
sw_clients: add a column for PyLink (clientbot)

GLolol:


DanielOaks:
I think this makes sense here. Thanks for the PR!

PR #181, opened 2017-12-13
TheLounge now supports draft/message-tags

MaxLeiter:


jwheare:
Just out of interest, how are tags being used in lounge? I couldn't find the commit that adds this cap.

jwheare:
Turns out this is inherited from irc-framework: https://github.com/kiwiirc/irc-framework/pull/117

PR #186, opened 2018-03-13
Adds Planio Team Chat

yeah:
I am the developer of Planio Team Chat and thought it would be helpful to add it to this list. Thanks for your consideration.


DanielOaks:
Thanks for the PR, much appreciated!

PR #187, opened 2018-03-17
Fix active link being pushed around

xPaw:
Just a minor nitpick, active link just ever so slightly moved when switching pages. This fixes that.


xPaw:
@DanielOaks Could you take a look at this?

PR #189, opened 2018-05-10
Update UnrealIRCd server capabilities

KoraggKnightWolf:
UnrealIRCd has support for CAP v.3.2, sasl-3.2, cap-notify, extended-join and chghost since version 4.0.16
These changes should be reflected in the listed/supported specs on the IRCv3 working group's site.

The source, with perhaps ones that were overlooked, can be obtained here: https://forums.unrealircd.org/viewtopic.php?f=1&t=8764

Thank you very much for your time and effort in advance.

Regards,

Koragg

### DIFF
--- a/_data/sw_servers.yml
+++ b/_data/sw_servers.yml
@@ -226,6 +226,7 @@
           cap-3.1:
           cap-3.2:
           sasl-3.1:
+          sasl-3.2:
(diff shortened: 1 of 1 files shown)


DanielOaks:
Noice. Thanks for the PR, much appreciated!

PR #190, opened 2018-07-14
Update sw_servers.yml to include weechat support updates

trobotham:
Signed-off-by: Terrance Robotham <nitemare@darenet.org>


jwheare:
Dupe of #191 with missing details

PR #191, opened 2018-07-14
sw_clients: update for WeeChat 2.2

sim642:


jwheare:
Looks good, need to check out that travis error though.

PR #201, opened 2018-11-05
BitBot now supports STS

jesopo:
Supported as of https://github.com/jesopo/bitbot/blob/061a4ede23abaf1c8cdbbd7a043161a980cdff77/modules/sts.py


jesopo:
As of https://github.com/jesopo/bitbot/blob/7408bfc9/modules/sts.py due to expiration rescheduling not being supported

PR #213, opened 2019-01-29
Add glirc to Client support list

glguy:
glirc is a Haskell based client that's a few years old and probably has dozens of regular users. The source tarball has been downloaded 10195 total times (511 in the last 30 days). It's not a mainstream client, but I make a point to try and support IRCv3 extensions, especially those used by ZNC or freenode.

* https://github.com/glguy/irc-core
* https://hackage.haskell.org/package/glirc


DanielOaks:
Noice, thanks for the PR!

PR #225, opened 2019-02-12
BitBot supports `echo-message`

jesopo:
As of jesopo/bitbot@b2d94202, bugfixed by jesopo/bitbot@2e925a68


RyanSquared:
No, my one advantage! 😦

PR #235, opened 2019-02-22
Update BitBot reference URLs

jesopo:
[BitBot](https://github.com/jesopo/bitbot) recently had his protocol parsing code split out in to different files - making the current link to `line_handler.py` invalid.


dequis:
In the future consider using a tag or a commit hash instead of 'master' in the url, so they are true permalinks and also include information of what version was looked up for the last update.

PR #236, opened 2019-02-24
KICL now supports message-tags

mbax:
Hurray ratification! :)


DanielOaks:
Nice! Thanks for the PR :)

PR #239, opened 2019-03-02
Fix Willow Chat participants URL

CarrotCodes:
The Willow Chat domain expired and is incredibly expensive to renew, so I've stuck the site back up using GitHub Pages.


DanielOaks:
Nice, thanks for the PR!

PR #242, opened 2019-04-05
Update Sopel's capability support

dgw:
Added version numbers from trawling back in the commit logs, and advertise forthcoming support for `echo-message` in the Git version.


DanielOaks:
Noice, thanks for the PR!

PR #243, opened 2019-04-09
Add the Standard Replies registry

DanielOaks:
This is the registry of codes used with the `FAIL`, `WARN`, and `NOTE` messages introduced [here:](https://github.com/ircv3/ircv3-specifications/pull/357) https://github.com/ircv3/ircv3-specifications/pull/357

This registry is very intentionally going to be flexible, going to be updated to take work-in-progress stuff into account, and going to match what implementers are proposing and using the Standard Replies messages for.


DanielOaks:
Cool, I'll keep this updated regularly

PR #244, opened 2019-04-10
Add initial metadata registry

DanielOaks:
Based on https://github.com/ircv3/ircv3-specifications/issues/336

Travis fails on this because the yaml checker's a bit silly and fails with emoji (i wil fix dat). Otherwise, this feels like what we want out of a metadata registry yeah?

### DIFF
--- a/_data/metadata_registry.yml
+++ b/_data/metadata_registry.yml
@@ -1,0 +1,28 @@
+# Metadata keys
+- name: Metadata Keys
+  values:
+    - key: avatar
+      description: Avatar that graphical clients can show alongside the user's name
+      format: URL with an optional `{size}` substitution denoting the size to load. `{size}` can be one of 128, 256, 512, 1024, or 2048
+      examples: https://example.com/avatar/{size}/asdf.jpg
+    - key: bot-url
+      description: >-
+        Lets bot runners provide a URL for more info on a bot user so they can be
+        identified as such in UI
+      format: Valid URL
+      examples: |-
+        https://example.com/bot-info.html
+    - key: display-name
+      description: >-
+        Alternative name to use instead of a nick for display purposes. Useful for gateways
+        to chat services that allow spaces and other characters in nicks. A nick is
+        required for standard protocol level stuff but can be less prominent in the UI
+      format: Any string
+      examples: |-
+        James Wheare
+        Posiden🦑
+    - key: status
+      description: Status text that can be shown for users without making them 'away'
+      format: Any string
+      examples: |-
+        Working from home
--- a/_data/metadata_registry.yml
+++ b/_data/metadata_registry.yml
@@ -5,24 +5,34 @@
       description: Avatar that graphical clients can show alongside the user's name
       format: URL with an optional `{size}` substitution denoting the size to load in pixels
       examples: https://example.com/avatar/{size}/asdf.jpg
-    - key: bot-url
+    - key: bot
       description: >-
-        Lets bot runners provide a URL for more info on a bot user so they can be
-        identified as such in UI
-      format: Valid URL
+        Lets bot runners provide the name of their bot software. Setting this key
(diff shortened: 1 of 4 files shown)


jwheare:
> one of 128, 256, 512, 1024, or 2048

No need to restrict this. I’m already using smaller ones.

DanielOaks:
Yeah it's a bit interesting. Are we just gonna allow people to use any number there? Do regular image/avatar hosting services that allow receiving a resized image usually support any number, only multiples of 2, etc?

jwheare:
I’ve seen hosts that accept any number (gravatar, IRCCloud) and some that round numbers to the nearest checkpoint (slack I think?) I’m sure others would just ignore unknown numbers. All cases fallback to showing an image at least.

DanielOaks:
Sounds good, I'll adjust it. If need be we can suggest requesting again with different sizes down the line or w/e

jwheare:
Cool. Might be worth mentioning that it’s a standard URI template that can be filled in with widely available libraries https://tools.ietf.org/html/rfc6570

_data/metadata_registry.yml
```diff
@@ -0,0 +1,28 @@
+# User Metadata
+- name: User Metadata
+  values:
+    - key: avatar
+      description: Avatar that graphical clients can show alongside the user's name
+      format: URL with an optional `{size}` substitution denoting the size to load in pixels
```
DarthGandalf: Is it height or width?

_data/metadata_registry.yml
```diff
@@ -0,0 +1,28 @@
...
+      examples: |-
+        https://example.com/bot-info.html
+    - key: display-name
+      description: >-
+        Alternative name to use instead of a nick for display purposes. Useful for gateways
+        to chat services that allow spaces and other characters in nicks. A nick is
+        required for standard protocol level stuff but can be less prominent in the UI
+      format: Any string
+      examples: |-
+        James Wheare
```
DarthGandalf: You can add another example with Cyrillic or Chinese characters.

_data/metadata_registry.yml
jwheare: Original issue stated "square dimension" so both.

registry.md
```diff
@@ -12,6 +12,9 @@ This page lists the tags, capabilities, commands, batches and metadata keys that
 {% for type in site.data.registry %}
 <a href="#{{ type.name | slugify }}">{{ type.name }}</a>
 {% endfor %}
+{% for type in site.data.metadata_registry %}
```
Zarthus: i might've read over it or it's already somewhere in the code, but a link to the spec(s) would probably be useful for reference

DanielOaks: there are no specs to link to right now, but when/if we do add them in the future I'll definitely add a column for 'em

_data/metadata_registry.yml
DanielOaks: Sounds good, will do

registry.md
jwheare: Well, you're linking to PRs in the fail codes section.

DanielOaks: When there's a PR that lays out in-detail how particular keys should act and be used, will definitely make sense to link it here.

_data/metadata_registry.yml
```diff
@@ -0,0 +1,30 @@
+# User Metadata
+- name: User Metadata
+  values:
+    - key: avatar
+      description: Avatar that graphical clients can show alongside the user's name
+      format: URL with an optional `{size}` substitution denoting the size to load in pixels
+      examples: https://example.com/avatar/{size}/asdf.jpg
+    - key: bot-url
```
jesopo: `bot-url` -> `homepage` + `bot`?

https://github.com/ircv3/ircv3-specifications/issues/336#issuecomment-493010547 ?

PR #247, opened 2019-04-23
Update IRCCloud support table, add bouncer

jwheare:


jwheare:
This doesn't have a bnc (as client) section because it's the same as the normal client list.

PR #251, opened 2019-05-15
Registry data tweaks

dgw:
Just fixing some stuff I noticed while poking around the registry page.

I kept this as two commits in case the more opinionated second commit (which goes beyond simple typo-fixing) is undesired. Happy to squash them together if wanted!


DanielOaks:
There's a couple changes in here that feel interesting, but on the whole an improvement. Thanks for the PR!

PR #252, opened 2019-05-16
[message-ids] Ratified, remove draft prefix and update support tables

jwheare:
I've not updated the tables for InspIRCd or Oragono since I'm not sure if they need version number qualifiers or if their support are in releases yet.

### DIFF
--- a/_data/registry.yml
+++ b/_data/registry.yml
@@ -200,7 +200,7 @@
       description: Allows clients to correlate requests with server responses
     - name: msgid
       specs: 
-        - message-ids
+        - msgid
       description: Provides a server supplied unique message ID
     - name: time
       specs:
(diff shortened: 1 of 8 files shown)


DanielOaks:
Noice. Mind also removing the `draft: true` from [specs.yml](https://github.com/ircv3/ircv3.github.io/blob/master/_data/specs.yml#L66)?

_data/registry.yml
```diff
@@ -200,7 +200,7 @@
       description: Allows clients to correlate requests with server responses
     - name: msgid
       specs: 
-        - msgid
+        - message-ids
```
DanielOaks: This particular change breaks the registry table (the reg table depends on `specs.yml` which uses the `msgid` spec name). Change this back to `msgid` and it'll build all fine.

DanielOaks:
Cool, once that change to fix the registry table's made this is good to merge in.

PR #255, opened 2019-05-30
Add server-time capability to quassel

justjanne:
## In Short

* Add server-time to the capability list of quassel with version "git"

## Background Information

Quassel supports server-time in the [current Git version](
https://github.com/quassel/quassel/commit/e38846f054ad1766f2e91992a57bbaffd33c7c06).

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -221,6 +221,7 @@
           sasl-3.1:
           sasl-3.2: 0.13+
           userhost-in-names: 0.13+
+          server-time: Git
         SASL:
           - external
           - plain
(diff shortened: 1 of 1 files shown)


_data/sw_clients.yml
```diff
@@ -221,6 +221,7 @@
           sasl-3.1:
           sasl-3.2: 0.13+
           userhost-in-names: 0.13+
+          server-time: Git
```
digitalcircuit: You should add `message-tags`, too, as it's supported and [`server-time` depends on it](https://ircv3.net/specs/extensions/message-tags.html#capabilities ).

Might be worthwhile keeping the list in alphabetical order, too.

PR #258, opened 2019-06-07
Moon-Moon: Update links, update support values

RyanSquared:
Changed update links to follow restructuring of IRCv3 support, as well
as update the support tables for the new caps and tags:

- message-tags
- msgid
- sasl-3.1
- sasl-3.2
- +draft/reply
- +draft/typing

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -747,12 +747,14 @@
           extended-join:
           invite-notify:
           message-tags:
+          msgid:
           multi-prefix:
           sasl-3.1:
           sasl-3.2:
           server-time:
           userhost-in-names:
           +draft/reply:
+          +draft/typing:
         SASL:
           - plain
           - external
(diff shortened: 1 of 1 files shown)


jesopo:
not `+draft/typing`?

https://github.com/wiseguiz/Moon-Moon/blob/470f522fff3db5dc44c96e1adbf20c13e06824b9/irc.moon#L185-L191

RyanSquared:
I'm not 100% positive that one's tested and ready to go yet. Once I do a bit more testing. I think right now the fail case will still tell the client it's typing.

RyanSquared:
It's not *as* good as I'd like - for instance, the bot can be running two things at once, and it might finish one operation before it's done with another. Not sure how to work on that. Might design a hook system later to deal with that.

RyanSquared:
Above comment made redundant as of https://github.com/wiseguiz/Moon-Moon/commit/8d5a74ff5a88d5ab3f34b9b87ae872a99e580e87

PR #264, opened 2019-07-03
Add non endorsement disclaimer to software intro

jwheare:


justjanne:
Could the endorsement disclaimer get a colorful background to make it stand out more? Similar to what the draft/ specs have currently?

jwheare:
It doesn't need to be in your face, just there IMO.

justjanne:
This is just my personal opinion, but I believe this disclaimer should be relatively in-your-face, to make it visible enough that it’s read before users actually scroll down or read the client list. After all, often enough this list is linked to users who are new to IRC and wish to install their first IRC client after having used qwebirc or Mibbit before.

digitalcircuit:
A less-aggressive tone that doesn't convey danger might work well, e.g. blue, as [on the LineageOS wiki pages](https://wiki.lineageos.org/devices/shamu ).

jwheare:
When determining whether this information should be in your face, consider: is it the most important piece of information on that page, or does it primarily exist so that it can be cited as an explicit policy and no one can credibly try to undermine the project by arguing that the list is an endorsement.

It's not really that valuable compared to the other information on the page for 99% of users, so does not justify stealing their attention.

justjanne:
@jwheare That page is mostly technical information, and IMO, shouldn’t be used by users to make decisions about which client to use.

The most important piece of information, in that case, is then that users shouldn’t use the page to make a decision about which client to use.

jwheare:
I don't mean IRC client users, I mean users of that page.

justjanne:
Well yes, that's my point — telling IRC users that this page isn't meant for them.

jwheare:
We disagree, so let's just wait for @DanielOaks to make a perfect edit and then push his version live with no further discussion.

justjanne:
I mean, this is just a minor thing, it's not like it's super important. Whatever ends up being chosen, it's gonna be an improvement over the alternatives either way.

DarthGandalf:
For me it's fine either way. But this PR is written already.

PR #267, opened 2019-07-13
Remove Mutter from Client List

AndroidKitKat:
Development of [Mutter](https://www.mutterirc.com/) has ceased and can no longer download from the App Store.


AndroidKitKat:
Removing stale software appears to be the precedent: https://github.com/ircv3/ircv3.github.io/pull/223

PR #275, opened 2019-09-25
update oragono caps

slingamn:
msgid support was released in 1.1.0:

https://github.com/oragono/oragono/blob/1b0e9a0c081852e435a1b760c8b47103c3b586fd/CHANGELOG.md#110---2019-06-27

and is present in the current stable release 1.1.1. Thanks!


slingamn:
(The same releases also include support for ratified message-tags.)

DanielOaks:
\o/

PR #192, opened 2018-08-08
Update Thunderbird

freaktechnik:
Thunderbird 60 was just released and added support for server-time.
Instantbird is no longer a maintained project: http://blog.queze.net/post/2017/10/18/Thunderbird-is-the-next-version-of-Instantbird

@clokep

PR #285, opened 2019-12-04
Update supported features in Thunderbird

clokep:
We just merged echo-message support into Thunderbird for version 73: https://bugzilla.mozilla.org/show_bug.cgi?id=1601102

I'm unsure why message-tags was never marked as supported, it was added at the same time as server-time support (in https://bugzilla.mozilla.org/show_bug.cgi?id=1302447)

I think #192 just missed that it was a separate entry in the table.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -200,7 +200,7 @@
           - external
           - plain
     - name: Mozilla Thunderbird
-      # ref: irc{CAP,MultiPrefix,SASL,ServerTime,WatchMonitor}.jsm files in
+      # ref: irc{CAP,EchoMessage,MultiPrefix,SASL,ServerTime,WatchMonitor}.jsm files in
       #      https://searchfox.org/comm-central/source/chat/protocols/irc/
       link: https://www.thunderbird.net/
       support:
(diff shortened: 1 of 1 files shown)


clokep:
CC @freaktechnik.

jwheare:
`message-tags` is a recentish CAP that clients can request now, it's not the same as just having protocol support for tags.

jwheare:
https://ircv3.net/specs/extensions/message-tags.html#message-tags-capability

clokep:
Ah, thanks. I suspect we just need to respond that it is supported then. @freaktechnik any thoughts?

jwheare:
At minimum you'd need to also handle the `TAGMSG` command, and client-only tags (prefixed with +) in order to request that cap, and make sure you're aware of the changed size limits on the tag space https://ircv3.net/specs/extensions/message-tags.html#size-limit

clokep:
@jwheare Neat. Thanks for the info. I've filed a [separate bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1601358) on our end to track this. For now I've rebased this branch to just be about echo-message.

PR #292, opened 2020-01-03
Update server software table for UnrealIRCd 5

syzop:
Update https://ircv3.net/software/servers.html with the IRCv3 features that UnrealIRCd 5 supports.


DanielOaks:
Awesome, thanks very much for the PR! \o/

PR #296, opened 2020-02-14
Add pounce bouncer to support tables

causal-agent:
(I am the author of [pounce](https://git.causal.agency/pounce/about/))

I added this between IRCCloud and ZNC assuming the list was sorted but I now see that KiwiBNC appears after ZNC, so not sure if there's an intended order.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -638,6 +638,42 @@
           server-time:
           sts:
           userhost-in-names:
+    - name: pounce (as Server)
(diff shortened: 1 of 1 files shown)


SadieCat:
I think KiwiBNC being listed after ZNC is a mistake and it should be alphabetical.

DanielOaks:
Honestly I need to fixup and alphabetize a lot of the tables. I'll merge this then do it~

PR #299, opened 2020-02-17
Add PIRC.pl network

k4bek4be:


jwheare:
Convention is currently not to include the metadata key since the spec is still in considerable flux and there is no published draft or non-deprecated spec listed on the website yet.

There are a couple of projects that do list it atm, but that should probably be fixed, and the site doesn't do anything with the key in support tables anyway.

PR #302, opened 2020-02-18
Ratify setname

jwheare:


jwheare:
Unprefixed support is in
* inspircd (git)
* bitbot
* irccloud

Still with draft prefix, later PRs can update these when they're ready:
* palaver
* thelounge
* irssi
* oragono

PR #308, opened 2020-03-04
Update support table for pounce 1.2

causal-agent:


_data/sw_clients.yml
```diff
@@ -634,37 +634,58 @@
...
+          labeled-response: 1.2+
+          message-tags: 1.2+
           monitor:
+          msgid: 1.2+
           multi-prefix:
           sasl-3.1:
+          sasl-3.2: 1.2+
+          server-time: 1.2+
+          setname: 1.2+
+          sts: N/A (TLS-only)
```
DanielOaks: This should be marked N/A separately, as it's done over here: https://github.com/ircv3/ircv3.github.io/blob/master/_data/sw_servers.yml#L17

DanielOaks:
Other than that N/A fix, nice! Once that's switched over we can merge this no worries.

_data/sw_clients.yml
causal-agent: ah, didn't see that that was a thing

PR #310, opened 2020-03-10
Enable link underlines

xPaw:
It's currently hard to see what's actually a link, this PR does a few things: enable link underlines in content and adds hover states.

I also tweaked the aria attributes for icons in the navbar.


DanielOaks:
[here](https://github.com/ircv3/ircv3.github.io/blob/8317fb81bcef85b4bde8daaf461c3ada2b395a1d/css/style.sass#L333) in `style.sass`, underneath the `colour: #DB1438;`, add these two properties:

```
text-decoration-color: #DB1438;
text-decoration: underline;
```
this makes the monospace'd links look much nicer

DanielOaks:
Nice, looks good to me

jwheare:
Yup seems good. Even though it's kind of ugly, the clarity is worth it.

Tiny thing, the orange RSS button on https://deploy-preview-310--ircv3.netlify.com/wg has inline styles, can you check a `text-decoration: none` on it?

PR #312, opened 2020-04-22
Add "color" metadata key information.

k4bek4be:


_data/metadata_registry.yml
```diff
@@ -14,6 +14,15 @@
         PacketBot v1.2
         JitiBot
         melo3
+    - key: color
+      description: >-
+        Denotes a color the user likes to be associated with. Client designers should decide,
+        which UI elements are using this color (for example avatar placeholders, nicknames
+        or whole otherwise-uncolored messages).
```
jwheare, line 21:
```suggestion
        A color to associate with a user. Client designers may decide which UI elements use this color (for example avatar placeholders, nicknames or whole messages).
```

DarthGandalf:
FTR, KVIrc supports specifying 2 colors - foreground and background, and that's used to show the nick in chat for other users.

But those colors are usual mIRC codes, not html colors. And it doesn't currently use metadata to communicate it, but something custom.

PR #317, opened 2020-05-06
update oragono testnet

slingamn:


slingamn:
Netlify preview looks as expected.

PR #322, opened 2020-05-20
CLIENTTAGDENY added, +typing ratified, account-notify update

jwheare:
* CLIENTTAGDENY: https://github.com/ircv3/ircv3-specifications/pull/412
* +typing: https://github.com/ircv3/ircv3-specifications/pull/411
* account-notify: https://github.com/ircv3/ircv3-specifications/pull/407

### DIFF
--- a/_data/irc_versions.yml
+++ b/_data/irc_versions.yml
@@ -164,6 +164,7 @@ stable:
       name: +typing
       description: Typing client tag
       link: /specs/client-tags/typing.html
+      hide-if-no-support: true
       tags:
         - +typing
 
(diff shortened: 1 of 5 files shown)


jwheare:
This is currently showing +typing on server/bouncer support tables. @DanielOaks will need to do a followup that hides them, I'm not sure on the best way with our data templating setup.

DanielOaks:
hmm, what do we want to display there instead? 'typing', 'typing c2c', something different? keeping it short as we can is good because otherwise things tend to overflow with the way we display tables

PR #339, opened 2020-12-17
Add soju to the support tables

hhirtz:
cc @emersion

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -818,6 +818,44 @@
           sasl-3.2:
           server-time:
           userhost-in-names:
+    - name: soju (as Server)
+      link: https://sr.ht/~emersion/soju
+      support:
+        stable:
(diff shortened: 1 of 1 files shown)


_data/sw_clients.yml
```diff
@@ -818,6 +818,39 @@
...
+          message-tags:
+          msgid:
+          multi-prefix:
+          sasl-3.1:
+          sasl-3.2:
+          server-time:
+    - name: soju (as Client)
+      link: https://sr.ht/~emersion/soju
+      support:
+        stable:
```
emersion, line 842: Could add a `SASL` entry below, as we support both PLAIN and EXTERNAL. See the entry for znc which also support both for instance.

hhirtz: Done!

DanielOaks:
Perfect, thanks very much for the PR!

PR #342, opened 2021-01-04
Update Snoonet spec support.

SadieCat:
All listed specs are supported on servers which are currently in the irc.snoonet.org DNS pool. Some older servers which are linked to the network but not in the DNS pool may not support all specs for a few weeks as we're waiting for them to drain so they can be upgraded.


DanielOaks:
Noice, thanks for the PR!

PR #345, opened 2021-01-25
tilde.chat updates

benharri:
- remove specific inspircd version number
- mark sts as partial support: we only allow TLS connections
  and have removed the STS module

### DIFF
--- a/_data/su_networks.yml
+++ b/_data/su_networks.yml
@@ -210,7 +210,7 @@
         stable:
           message-tags: No client tags
     - name: tilde.chat
-      ircd-ver: inspircd-3.6.0
+      ircd-ver: inspircd-3
       net-address:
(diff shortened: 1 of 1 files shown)


SadieCat:
>mark sts as partial support: we only allow TLS connections and have removed the STS module

If you've removed it entirely it should be n/a not partial

jwheare:
It shouldn't be listed at all if it's been removed.

benharri:
is there a way to have it appear red with N/A instead of "no"?

jwheare:
Well, STS *is* still applicable even if you only allow TLS connections. It prevents clients ever succumbing to a mitm attack on the plain port.

benharri:
We do allow plaintext connections on loopback on some shell servers - but I'll have a look at setting up another listener for external connections with STS only.

PR #351, opened 2021-02-14
add pinkieval to Technical Board

jesopo:
I imagine that it would be better for someone from the current technical board to submit this PR, but I imagine it doesn't practically matter and if not me, who?

pinkieval's a great contributor to spec drafting and has enthusiastically implemented a lot of new things in a heavy hitting popular IRC bot. obvious choice!

### DIFF
--- a/_data/tc.yml
+++ b/_data/tc.yml
@@ -32,6 +32,11 @@
     - "[IRCCloud](https://www.irccloud.com/)"
   chair: true
 
+- nick: pinkieval
+  github: ProgVal
+  projects:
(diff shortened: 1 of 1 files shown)


ProgVal:
Thanks for the nomination.

I didn't expect it, but I'd be happy to join. If the TB wants to have me, of course.

jwheare:
I'm for it.

justjanne:
I concur, pinkieval seems like a good fit bringing a useful perspective onto the board.

PR #355, opened 2021-02-28
Spec cleanup LOOKIE EXAMPLE

DanielOaks:
Yeah this is the spec cleanups in https://github.com/ircv3/ircv3-specifications/pull/441 but in here so we can have a look at it when Netlify builds it. **Don't merge this PR,** it will do weird stuff because of where I've applied the `specs` submodule changes from.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,7 @@
 name: IRCv3 Working Group
 markdown: kramdown
 highlighter: rouge
-url: https://deploy-preview-355--ircv3.netlify.app/
+url: https://deploy-preview-355--ircv3.netlify.app
 repository: ircv3/ircv3.github.io
 kramdown:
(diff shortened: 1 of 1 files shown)


jwheare:
Need to update URLs/fullnames in _data/specs.yml too.

jwheare:
And _data/irc_versions.yml

DanielOaks:
Sounds good, will do on the real PR (note to self, consider whether to change the `starttls` name in those files to `tls` considering that's what the cap and spec is called. I think I used `starttls` here to disambiguate between deprecated starttls and general like, TLS, the protocol though, so may just leave it).

jwheare:
LGTM

DanielOaks:
I can close this now, this is merged proper into the site

PR #362, opened 2021-03-26
Generate header ids and anchors without Javascript

ProgVal:
Uses https://github.com/allejo/jekyll-anchor-headings instead of
https://www.npmjs.com/package/anchor-js

As far as I can tell, it only breaks one anchor:
https://ircv3.net/charter#project-resources-contribution
will become https://ircv3.net/charter#project-resources--contribution
because kramdown does not generate ids in exactly the same way as
anchor.js

Pros of this change:

* Works in browsers without JS
* Less code overall (and the code is readable instead of being minified)
* No need for `{% include anchors.html %}` at the bottom of every page

Cons:

* Breaks that URL
* A little more work on the server side


jwheare:
That broken link isn't a big deal. I'd prefer if the styling could be kept though. Link icon appears on hover, not just an always visible # on every header.

ProgVal:
> I'd prefer if the styling could be kept though.

Do you mean the icon instead of `#`?

> Link icon appears on hover, not just an always visible # on every header.

That what my change to the `.sass` does.

ProgVal:
(updated styling to look like the old one, but using an emoji instead of a base64ed image)

jwheare:
The emoji is a bit of an eyesore on windows. Image is probably better.

ProgVal:
Done.
    
I chose an an image from emojitwo instead of using the same one as anchor.js,
because anchor.js' comes from icomoon.io, which seems to have only
commercial licenses

DanielOaks:
Give me a lil bit, gonna look at the support tables and fix their generation with this.

DanielOaks:
Nice, after that change it seems good. I'm happy enough with that new link logo, but we can always search for a better one down the line. Thanks for the PR!

PR #366, opened 2021-05-02
Update the client specification support list. 

SadieCat:
- Update client spec support
- Remove/replace some dead clients

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -33,7 +33,7 @@
           plain:
           scram-sha-256: 3.4+
     - name: Ambassador
-      # ref: JSIRCV3_SUPPORTED_CAPS in https://github.com/Ascrod/ambassador/blob/master/ambassador/js/lib/irc.js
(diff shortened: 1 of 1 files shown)


ProgVal:
Could you use a tag or commit id in links, instead of `master`? Links to `master` may break because files or code is moved.

SadieCat:
Tag comments are kinda useless for referencing because when you go to update the list it just points to an old commit and you have to find the branch anyway.

ProgVal:
but it's useful to provide a source for the existing data. And if someone wants to see the latest version, they can replace with `master` in the URL

_data/sw_clients.yml
```diff
@@ -397,6 +430,7 @@
           - plain
     - name: The Lounge
       # ref: https://github.com/thelounge/thelounge/projects/1
+      #      https://github.com/kiwiirc/irc-framework/blob/master/docs/ircv3.md
```
ProgVal, line 432: wrong link

SadieCat: It's not the wrong link. The Lounge uses irc-framework.

ProgVal: ok, my bad

DanielOaks:
Thanks for the PR, much appreciated! <3

PR #367, opened 2021-05-02
Update the server and network lists for the latest spec support. 

SadieCat:


DanielOaks:
Mm. Coolcool. Let's see whether anyone comes to bat for Chary or asks us to put it back on the table, but yeah I'm okay with this for now.

PR #370, opened 2021-05-22
Fix checknetworks script

DanielOaks:
`girc` for some reason stopped working, wrote this lil reliable connection script instead. PR because I wanna confirm the n/w updates look fine.


DanielOaks:
Honestly this needs to read `irc_versions.yml` and get the caps/isupports from there to check. And fix that small issue with it not picking up partial cap supports properly. But this is fine for now and I can commit that on top of master.

PR #371, opened 2021-05-23
Update the Konversation spec support.

SadieCat:
Somehow I missed this when doing the list.


DanielOaks:
(that yaml validate failure's from me earlier, please ignore it)

PR #373, opened 2021-05-24
Update the client lists for the latest changes in HexChat Git.

SadieCat:


TingPing:
`message-tags` also.

TingPing:
Oh nevermind, there is actually a cap for `message-tags` which isn't supported.

DanielOaks:
Thanks for the PR!

PR #378, opened 2021-06-02
Add Eggdrop to Bots

vanosg:


_data/sw_clients.yml
```diff
@@ -924,6 +924,26 @@
           - plain
           - external
           - scram-sha-256
+    - name: Eggdrop
+      link: https://github.com/eggheads/eggdrop
```
DarthGandalf, line 928: Can you provide any reference link in source?

vanosg: Sure, sorry, I wasn't sure of the intent of that line- what is this intended to reference? If code, it's kinda smashed all throughout. Maybe a reference to the config file?

ProgVal: You can link to the list of capabilities it requests, and add an extra link for the few extensions with no capabilities

DarthGandalf:
Thanks!

PR #379, opened 2021-06-02
Add gamja to support tables

emersion:


DanielOaks:
Nice, thanks for the PR! @emersion would you be able to add this to the start of the web clients list? (alphabetical, etc).

DanielOaks:
Nice, thanks mate!

PR #380, opened 2021-06-02
Split 'Desktop clients' into GUI/TUI and make tables less wide

progval:
So they fit on 13' screens without overflowing.

I also think it makes sense to split clients this way,
as they don't have the same audience.

Rendered view

### DIFF
--- a/_includes/support_list.html
+++ b/_includes/support_list.html
@@ -18,8 +18,8 @@
 {% assign sw_index = 0 %}
 {% assign sw_index_body = 0 %}
 {% assign sw_count = 0 %}
-{% assign max_sw_per_table = 9 %}
-{% assign min_fullwidth_display = 8 %}
+{% assign max_sw_per_table = 7 %}
+{% assign min_fullwidth_display = 6 %}
 
 {% for abcxyz in (1..9999) %}<!-- we expect to break out of this loop in all cases, so this is fine -->
 {% assign sw_count = 0 %}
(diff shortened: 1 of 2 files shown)


DanielOaks:
Thanks for the PR!

Hmm, could we name them Desktop Clients and Terminal Clients, instead of Graphical Desktop and Console Desktop?

They may not have the same audience but I think it makes sense for them to sit in the same spot, just underneath Desktop Clients (so it'd be Desktop Clients, Terminal Clients, and then Web Clients).

I think the specs/ submodule accidentally got changed here as well, could you revert that to just being the one master's on? (currently `1b4050e`)

ProgVal:
Sure, done.

_includes/support_list.html
```diff
@@ -18,7 +18,7 @@
 {% assign sw_index = 0 %}
 {% assign sw_index_body = 0 %}
 {% assign sw_count = 0 %}
-{% assign max_sw_per_table = 9 %}
+{% assign max_sw_per_table = 7 %}
```
DanielOaks, line 21: Can you also update `min_fullwidth_display` in the next line to be `6`?

ProgVal: Good catch, that's much better.

DanielOaks:
I'll leave this PR here for now. With how things get pushed down with the tables currently, it's hard to justify this for what we get in logical separation. I'm gonna open an issue for switching the columns+rows of the table and some other simplifications, and then we can re-evaluate this split when it won't cause as much hiding of other clients.

progval:
actually, this PR isn't relevant anymore

PR #381, opened 2021-06-02
Add OS icons for graphical desktop clients

progval:
This doesn't add icons to Terminal Clients, so this PR does not conflict with #380

I'll add them if we decide to merge this PR before #380.

Note that the "freebsd" icon isn't displayed because we use FontAwesome 4, and it was added in FontAwesome 5.


DanielOaks:
I don't mind this but let me look into the table changes first (it's a tiny bit unwieldy with the current layout). I may split this into a separate column depending on stickyness of the top/left headers

ProgVal:
Sure, no hurry

DanielOaks:
We'll need to update FontAwesome to get the bsd icon but we can worry about that later.

For now could you add this to `_swtable.scss` just under [where the other os support colours are](https://github.com/ircv3/ircv3.github.io/blob/master/css/partials/_swtable.scss#L328):
```css
.os-support-macos {
    color: #444;
}
.os-support-windows {
    color: #4d75b3;
}
.os-support-linux {
    color: #826d2e;
}
.os-support-bsd {
    color: #a42f2f;
}
```

bitchx and swirc both support all of the OSes, according to their websites/install instructions.

I think Glirc supports all of them as well?

SadieCat:
It's probably fine to just group BSD and Linux under a general "UNIX" category.

jwheare:
Old

progval:
but still relevant. will it be merged if I do it again?

jwheare:
Oh I assumed it was no longer valid. I'll reopen.

PR #383, opened 2021-06-03
Oragono is now Ergo

slingamn:
We're starting a production network in addition to the testnet, but I'll leave that out for now since we haven't finalized all the policy decisions.

### DIFF
--- a/_data/su_networks.yml
+++ b/_data/su_networks.yml
@@ -2,6 +2,41 @@
   note: >
     These networks are intended to test new IRCds and can also be used to test new client implementations of IRCv3 specifications.
   software:
+    - name: Ergo Testnet
+      ircd-ver: master
+      net-address:
+        link: ircs://testnet.ergo.chat:6697/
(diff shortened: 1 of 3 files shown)


_data/su_networks.yml
```diff
@@ -65,12 +65,12 @@
       na:
         stable:
           starttls: direct TLS only
-    - name: Oragono Testnet
+    - name: Ergo Testnet
```
DanielOaks, line 68: Could you please move this to the start of the testnet list? (alphabetical order, etc)

_data/sw_servers.yml
```diff
@@ -105,9 +105,9 @@
           starttls:
           userhost-in-names:
           webirc:
-    - name: Oragono
-      # ref: https://oragono.io/specs.html
-      link: https://oragono.io/
+    - name: Ergo
```
DanielOaks, line 108: If you could move this so that it's just after Bitlbee that'd sort out the alphabetical order

PR #388, opened 2021-06-15
HexChat supports the UTF8-only spec in Git.

SadieCat:


jwheare:
How should clients that only deal with UTF8 be treated in the support tables for this?

DanielOaks:
We could mark them N/A or green. Maybe Green with the text 'UTF8 only' would make the most sense?

ProgVal:
What about green with "Always" as text?

PR #389, opened 2021-06-15
Remove freenode from the IRCv3 networks list.

SadieCat:
- Freenode has recently migrated to a new IRCd so this is no longer valid.
- Consent for it to be listed here was provided by the previous staff team not the the current staff team.
- Freenode staff now also includes someone who is banned from the IRCv3 project for abusing contributors so it feels inappropriate to include it here.

### DIFF
--- a/_data/su_networks.yml
+++ b/_data/su_networks.yml
@@ -156,24 +156,6 @@
           sasl-3.1:
           starttls:
           userhost-in-names:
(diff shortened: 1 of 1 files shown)


justjanne:
Freenode has modified their new IRCd as well, so it's unclear how compliant their network or bnc are with the specifications.

PR #394, opened 2021-07-11
Fix yaml validation

ProgVal:
`net-address.exit-early` is a boolean, and setting `net-address` as
`mapofstring` validated it incorrectly.


DanielOaks:
Oh my god. Thank you so much, I could not for the life of me work this out after staring at it. <3

Issue #350, opened 2021-02-14
Technical board membership and processes


DanielOaks:
The technical board's stayed largely the same for a fair while now. So have the processes that IRCv3 has worked with. Figured I'd make an issue to discuss that (just opening it in this repo because this is where previous TB membership has been discussed in the past).

---

So, what other projects make sense to bring onto the TB? Or are there any irc-related people who've been doing stuff actively that people think we should bring on?

Obviously, Atheme (iirc last time we chatted to them we couldn't come to agreement on who to be put forward to represent them), networks are always useful so maybe someone representing solanum and/or one of the networks that're putting work into that project? They're both really big, obvious hitters, who else has been active and pushing development?

In general, more network involvement would be appreciated I think - but aside from freenode a lot of the other big networks seemed not-that-interested in working with us because of previous bad experiences in years past. Should we reach out again? If so, who should we reach out to.

I know that I've been really inactive, and sling's been doing more than me in terms of pushing stuff towards v3?

---

Somewhat of a side discussion, given the lack of real 'power' that TB membership brings (there's the perceived power of course, but in terms of process it's more limited and I think we tend to be pretty lax in terms of most parts of the process that require 'actual authority'), I find that most of the time when people ask for TB changes what people are really looking for is a change in the processes the group runs with.

In https://github.com/ircv3/ircv3-specifications/pull/375 (or along with when I was writing up that PR) I was really messing about with changing up our processes in general, overhauling them to make them more defined or to try and speed up useful, active development. Lately I've been a lot more bent on the idea of implementors getting together to build ideas, prove they work, and then submitting them to us once they're ready and running (ala znc with server-time, irccloud with the client-tags they introduced, etc). Anyone have ideas about changing processes or ideas to improve them, or itches/issues they have with the current processes v3 works under?

jesopo:
I'd like to suggest we look at reinstating meetings with set times and agendas

DanielOaks:
Regarding processes, https://github.com/ircv3/ircv3-specifications/pull/215 and https://github.com/ircv3/ircv3-specifications/issues/214 are I think the last places we explicitly discussed our processes around things getting ratified.

DanielOaks:
[...]

PR #71, opened 2016-02-23
New charter and code of conduct

jwheare:
This pull request establishes the proposals from ircv3/ircv3-specifications#233 by updating the charter [[rendered](https://github.com/jwheare/ircv3.github.io/blob/charter-coc-update/charter.md)] and adding a new code of conduct [[rendered](https://github.com/jwheare/ircv3.github.io/blob/charter-coc-update/conduct.md)].

The "Community Council" is renamed to "moderation team" and "BDFN" is renamed to "chair" of the technical board.

The old "Ground rules" are incorporated into the code of conduct, adapted from the latest version of the [Contributor Covenant](http://contributor-covenant.org/) [[1.4](http://contributor-covenant.org/version/1/4/)] [[diff](https://gist.github.com/jwheare/df473e0b24d845e0187d)] that more explicitly spells out unacceptable behaviour, and specifies how abuse is reported and enforced. The moderation team is listed there.

Other documents I researched when putting this together include:
* https://www.npmjs.com/policies/conduct
* https://golang.org/conduct
* https://www.rust-lang.org/conduct.html
* http://hood.ie/code-of-conduct/
* http://citizencodeofconduct.org/

Followups:

* [x] I'd like to set up a <strike>community@ircv3.net</strike> ircv3-mods@googlegroups.com email address that forwards to the moderation team. I'll talk to the relevant people in IRC about this.
* [ ] We need more moderators, if anything to cover more time zones. They need to be people who are prepared to enforce the CoC. Does anyone want to volunteer or nominate someone?
* [ ] Close ircv3/ircv3-specifications#222 and ircv3/ircv3-specifications#233
* [ ] Finalise and merge ircv3/ircv3-specifications#215 adding appropriate amendments to the charter.
* [x] Address #67, either closing it in favour of the technical board listing, or revising it to list "members" rather than "stakeholders"
* [ ] I will be speaking to other projects not currently listed on the technical board to see if we can get a more well-rounded list.

PR #399, opened 2021-08-06
Deprecate Technical Board

DanielOaks:
The 'Technical Board' as a concept has done a lot more harm than good to v3, particularly in recent months.

We made the TB to balance out a tyrannical chair and basically have some formal way to remove and replace the chair if we ever needed to. Well, we luckily haven't had to, and if we did need to in future we could probably do so without having 'The Technical Board' as a tangible thing (especially given their utter lack of any real power right now).

The current TB has zero power. No voting, no exclusive chatting/submission/etc rights, literally nothing. But the amount of effort we need to expend going through the 'voting' process to add new people to the list, to argue about whether or not someone should go onto the list, and to argue about how many people from each project should be allowed on the list, is mind-boggling.

For more info on why the TB exists as it does today, see these issues:

- https://github.com/ircv3/ircv3-specifications/issues/222
- https://github.com/ircv3/ircv3-specifications/issues/233
- https://github.com/ircv3/ircv3.github.io/pull/71
- https://github.com/ircv3/ircv3.github.io/issues/350

-----

This change replaces the Technical Board list with an informal list of Contributors. It's acknowledged that the list here has no real power and is just to show off and shout out some of the people who contribute to the IRCv3 Working Group.

If people want to be added to the new Contributors list it should be a simple matter of confirming they've been contributing, have been playing nice in the WG / they haven't violated our code of conduct, their details are correct, and then hitting Merge on the PR. It doesn't need to be any more complicated than that.

-----

As far as I'm aware, this roughly matches the wishes of a large part of the v3 group (including the 'technical board') and when I brought doing this up in the past I didn't get any large amount of pushback from the group.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -14,23 +14,19 @@ in both areas.
 These extensions are published on this website through backwards-compatible [specifications]({{site.baseurl}}/irc/).
 
 
-## Technical board
+## IRCv3 Contributors
 
-The direction of these specifications is led by the technical board. The board consists of a broad range of IRC software and specification authors that reflect the community at large. The board will seek to make all decisions by consensus.
+The direction of these specifications is led by our contributors and the WG Chair. Our contributors consist of a broad range of IRC software and specification authors that reflects the IRC development community at large. We seek to make all decisions by consensus.
 
-Nominations for new members of the board will be considered once the following pre-requisites are met:
+To be listed on this page, contributors must:
 
-- Demonstrated support of the IRCv3 specifications
-- Approval from the existing board
-- Adherence to the [code of conduct]({{site.baseurl}}/conduct.html)
+- Demonstrate support of the IRCv3 specifications
+- Have the approval of the current contributors
+- Adhere to our [code of conduct]({{site.baseurl}}/conduct.html)
 
-In general, candidates who meet any of the following criteria will be given more consideration
+We're seeking representation from as many members of the IRC community as possible. If you or your project would be a good candidate for representation, please contact us!
 
-- Developers of widely used IRC software
-- Administrators of widely used IRC networks
-- People who have otherwise contributed materially to the specifications
-
-We're seeking representation from as many members of the IRC community as possible. If you or your project would be a good candidate for representation, please contact the board. The current members of the board are:
+Here's some of our current contributors:
 
 <table>
     <thead>
@@ -60,7 +56,7 @@ We're seeking representation from as many members of the IRC community as possib
     </tbody>
 </table>
 
-The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus isn't possible, including returning proposals to the drawing board where necessary.
+The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus between contributors isn't possible, including returning proposals to the drawing board where necessary.
 
 ## Project resources & Contribution
 
--- a/charter.md
+++ b/charter.md
@@ -56,7 +56,15 @@ Here's some of our current contributors:
     </tbody>
 </table>
 
-The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus between contributors isn't possible, including returning proposals to the drawing board where necessary.
+
+## Governance
+
+The WG Chair is someone the IRCv3 contributors have designated as the Leader For Now of IRCv3. They're able to move discussions forward and make decisions in cases where consensus between contributors isn't possible. They can also return proposals to the drawing board where necessary.
+
+In extraordinary circumstances, the IRCv3 contributors may decide to depose the WG Chair and replace them with someone else (as we've done [in the past](https://github.com/ircv3/ircv3-specifications/issues/233)). In addition, the Chair may appoint someone else to replace them as Chair.
+
+Ownership of the resources of the IRCv3 WG (including the GitHub org, domains, channel, and moderators email list) are spread across several contributors, to prevent the situation where someone can unilaterally control the WG without support from the contributors.
+
 
 ## Project resources & Contribution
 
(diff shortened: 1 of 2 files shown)


DarthGandalf:
As a member of the technical board, I approve this. Probably the only power I have as one.

SadieCat:
The original purpose of the technical board was to avoid the issue of the chair being an dictator for life in response to the drama that happened during the IRCv3.2 ratification. Whilst I have no complaints about the current chair (<3 @jwheare) maybe we should clarify this in the charter?

DanielOaks:
How's that change sound? I think this gives a reasonable overview of the 'Chair's' role, prior art on deposing dodgy ones (<3 u jw), and a clarification on the ownership of WG resources that's designed to prevent that sort of situation.

charter.md
```diff
@@ -60,7 +56,15 @@ We're seeking representation from as many members of the IRC community as possib
...
 
-The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus isn't possible, including returning proposals to the drawing board where necessary.
+
+## Governance
+
+The WG Chair is someone the IRCv3 contributors have designated as the Leader For Now of IRCv3. They're able to move discussions forward and make decisions in cases where consensus between contributors isn't possible. They can also return proposals to the drawing board where necessary.
+
+In extraordinary circumstances, the IRCv3 contributors may decide to depose the WG Chair and replace them with someone else (as we've done [in the past](https://github.com/ircv3/ircv3-specifications/issues/233)). In addition, the Chair may appoint someone else to replace them as Chair.
+
+Ownership of the resources of the IRCv3 WG (including the GitHub org, domains, channel, and moderators email list) are spread across several contributors, to prevent the situation where someone can unilaterally control the WG without support from the contributors.
```
DarthGandalf, line 66: Probably it should specify who controls what resource. Currently @jwheare controls everything except for ircv3.net domain, right?

charter.md
```diff
@@ -60,7 +56,15 @@ We're seeking representation from as many members of the IRC community as possib
     </tbody>
 </table>
 
-The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus isn't possible, including returning proposals to the drawing board where necessary.
+
+## Governance
+
+The WG Chair is someone the IRCv3 contributors have designated as the Leader For Now of IRCv3. They're able to move discussions forward and make decisions in cases where consensus between contributors isn't possible. They can also return proposals to the drawing board where necessary.
+
+In extraordinary circumstances, the IRCv3 contributors may decide to depose the WG Chair and replace them with someone else (as we've done [in the past](https://github.com/ircv3/ircv3-specifications/issues/233)). In addition, the Chair may appoint someone else to replace them as Chair.
```
DarthGandalf, line 64: This reads very weird. And vague. The point of charter/governance is specifically to avoid this kind of situation

slingamn:
I preferred the first revision of this:

1. I don't think we have a good sense of how to create a real governance structure around IRCv3, the way Libera (for example) has done successfully
2. I don't think there's a pressing need to do this; it seems like a distraction from the original motivating concern, which is that appointments to the technical board have become a divisive issue

If the goal is just to clarify that the chair doesn't have dictatorial powers, I think something like this would work:

>The WG Chair's authority is derived from, and contingent upon, the confidence that the community of contributors places in them.

DanielOaks:
How's something like this sound?

DanielOaks:
I think this list is accurate, let me know if there's any mistakes or resources I haven't listed

| Resource | Owner(s) |
| -------- | ------ |
| `ircv3.net` domain | DarthGandalf |
| `ircv3.org` domain | jwheare |
| `#ircv3` channel | dan-, jwheare |
| @IRCv3 Twitter account | jwheare |
| GitHub org | DarthGandalf, jwheare |
| ircv3-mods List | jwheare |
| Netlify account | DarthGaldalf |

charter.md
DanielOaks: Reworded to remove this section

slingamn:
Looks good to me, thanks for doing this!

DanielOaks:
> The original purpose of the technical board was to avoid the issue of the chair being a dictator for life in response to the drama that happened during the IRCv3.2 ratification. Whilst I have no complaints about the current chair (<3 @jwheare) maybe we should clarify this in the charter instead of just abolishing it outright?

So I think this is definitely still a valid concern to have. Something to properly reign in the power of the Chair, and to try to prevent those IRCv3.2 sorta issues from happening again. Let me lay out my thoughts on this proposal (keeping the TB and just clarifying that it's about avoiding the chair being a dictator).

-----

The term 'Technical Board' in the WG has come to mean (to those not on the board) a group of trusted contributors who have more power and flexibility within IRCv3. Being on The Board is desirable because it 'sets the direction of IRCv3 and the specifications we pursue', presumably while others in the WG don't have as much of an impact on that. And it makes sense for people to view the board that way, because that's literally how it's outlined right now on the website (despite it literally not being true. The text in this PR is a lot more accurate to how v3 runs today).

I don't think that we can retrofit the Technical Board to stay the same in terms of membership and name and just clarify their original purpose. The name 'Technical Board' itself heavily implies that they'll control the technical direction of v3, so even if we remove that sentence from the charter page the name of the group itself is pushing that as its purpose. We'll end up in the same situation with a bunch of hard-to-decide appointments to the TB and PRs that will continue the almost agonising membership discussions that have been plaguing us lately.

-----

To be entirely fair, this PR does remove a useful 'safeguard' against the chair being a dictator. But with the current chair and the ownership of project resources, I don't think that's a critical concern we need to address right this second. This PR also removes a major point of drama and something that has impacted how much motivation people have for doing useful work in the group, and I think that's more critical right now than ensuring that governance question is completely answered.

There are a lot of ways to handle those governance questions - setting up a way that the group can vote on a new chair, incorporating as a proper non-profit, making another table of names of people who can reign the chair in if another 3.2 ratification incident happens again, etc... But we can worry about that and chat more about that and debate the different solutions later I feel -- in the meantime, with the chair we have now, the line _"The WG Chair's authority is derived from, and contingent upon, the confidence that the community of contributors places in them."_ and the split of resource ownership should be more than sufficient to handle any issues that arise.

PR #409, opened 2021-10-03
Add WebSockets to support tables

DanielOaks:
Listed support for Ergo, InspIRCd, and UnrealIRCd.

I think this is correct? I know Kiwi and Gamja can both connect to Ergo's websockets support, but not sure if that's intentional support or just a byproduct of how the spec's written.

If any other software supports the WS spec, give me a shout in this PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,14 @@ You should make sure your manually installed local versions match that of [GitHu
 5. `git submodule update --init --recursive`
 6. `jekyll serve`
 
+## Adding New Spec
+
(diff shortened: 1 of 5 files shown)


emersion:
gamja hasn't merged support yet. soju only supports text.ircv3.net without binary.ircv3.net which makes it explicitly not spec-compliant.

slingamn:
Kiwi is working on a patch to comply with the spec, but the current master branch is non-compliant.

DanielOaks:
I'll merge this as-is for now

PR #437, opened 2022-03-04
networks: IRC4Fun support update

siniStar7boy:
updated the support table for IRC4Fun


jwheare:
Can you fix the conflicts here? Looks like you branched from an older version of this file.

PR #442, opened 2022-03-24
Merge _data/specs.yml into _data/irc_versions.yml

progval:
Easier to manage + remove inconsistencies in the names used to refer to specs
(often their cap-based names, sometimes a human-readable name -> now it's
always their cap-based name)


DanielOaks:
Oo, very nice! Thanks for tackling this work! Let me review this and dig into it, the data files here are... a mess, as you've probably seen.

DanielOaks:
Superseded by https://github.com/ircv3/ircv3.github.io/pull/447

PR #444, opened 2022-04-03
support lists: Remove vertical overflow+scrollbar + add footer

progval:
The overflow+scrollbar makes tables look weird, and I find them hard to use.

This is because vertical scrollbar within a webpage are not very
natural, but we needed them for headers to remain in sight.

With the header repeated at the bottom, this is no longer needed.
Not to mention the header is sticky when scrolling down the page anyway.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -280,4 +280,4 @@ DEPENDENCIES
   github-pages
 
 BUNDLED WITH
-   1.17.2
+   2.2.5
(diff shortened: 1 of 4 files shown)


jwheare:
I don't think we have a vertical scrollbar with sticky headers anymore, so this would just be for adding the footer which I personally don't think is that necessary (even on the longest list: Desktop Clients) and definitely should be hidden in the current mobile/small screen view.

jwheare:
Closed in favour of #451

PR #448, opened 2022-05-01
Add in support for account-registration to UnrealIRCd

ValwareIRC:
draft/account-registration is now support by UnrealIRCd via 3rd Party module.

### DIFF
--- a/_data/sw_servers.yml
+++ b/_data/sw_servers.yml
@@ -219,7 +219,6 @@
           cap-3.2:
           cap-notify:
           chghost:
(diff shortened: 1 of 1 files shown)


progval:
support via third-party modules should be in the "partial" section. Could you also make the description shorter?

Something like this: https://github.com/ircv3/ircv3.github.io/commit/9c8aa75114406c053ff67e1bf181f44de999e463

ValwareIRC:
like that?

progval:
yup

PR #451, opened 2022-05-02
Allow support table to break outside the container max width

jwheare:
This prevents punishing people with bigger screens

### DIFF
--- a/css/partials/_swtable.scss
+++ b/css/partials/_swtable.scss
@@ -20,6 +20,9 @@
 }
 
 .support-table {
+  width: 96vw;
+  position: relative;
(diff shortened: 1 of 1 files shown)


jwheare:
Compare:

https://deploy-preview-451--ircv3.netlify.app/software/clients

pageshot of 'Clients - IRCv3' @ 2022-05-02-1014'58

VS https://ircv3.net/software/clients

pageshot of 'Clients - IRCv3' @ 2022-05-02-1014'13

jwheare:
This supercedes #444 (but doesn't add the footer, I don't think it's necessary)

PR #459, opened 2022-07-01
Add draft/channel-context client tag + draft/read-marker extension

emersion:


progval:
Could you add `channel-context-client-tag` to Limnoria too?

progval:
thanks!

LGTM

_data/sw_servers.yml
```diff
@@ -219,6 +220,7 @@
           cap-3.1:
           cap-3.2:
           cap-notify:
+          channel-context-client-tag:
```
progval, line 223: I don't think it makes sense to mention it, except for Unreal it's redundant with message-tags. (and Unreal's also does not mention typing/react/reply)

PR #463, opened 2022-07-18
Update support table for mIRC 7.69

progval:
ref: https://forums.mirc.com/ubbthreads.php/topics/270512/mirc-7-69-released


jwheare:
Can probably drop the 7.62+ bits too.

PR #473, opened 2022-11-13
Added BonoboNET

deavmi:


SadieCat:
Do you have any servers which are accessible without using Tor/I2P/etc? I'm not sure we have any rules about this but it doesn't seem useful to include servers that most people can't access.

deavmi:
We have several IPv4/IPv6 servers accessible through the clearnet:

1. `41.157.98.109`/`2a04:5b81:2011::2`
2. `160.20.145.209`/`2a01:367:c1f2::254`
3. `2a04:5b81:2010::92`

deavmi:
Would you like me to add one of the above to the list?

deavmi:
@SadieCat ?

deavmi:
Many thanks good sirs

PR #476, opened 2022-11-28
Update the supported specifications for desktop clients.

SadieCat:


progval:
Could you keep referencing tags and commit ids instead of branches? We discussed it in #366

SadieCat:
Is this blocked on anything?

PR #479, opened 2023-01-08
sw_clients: swirc: added sasl mechanism 'external'

uhlin:
Starting from Swirc v3.3.9 the SASL auth mechanism EXTERNAL has been added.


jwheare:
Netlify failed preview on this for some reason but looks ok.

PR #487, opened 2023-02-16
Update Fontawesome and add Fediverse links

progval:
The FA version bump finally gives us the BSD icon we need for #381 :)


emersion:
It seems like the header bar icons on the right are broken in the preview.

screenshot

progval:
sigh, I forgot to grep in `_includes`

PR #490, opened 2023-02-21
Add dedicated page for bridges 

progval:
Purely because I want to show off the great draft spec support in https://github.com/progval/matrix2051 :)

@jwheare Feel free to add your Slack gateway, you have push access to the branch

### DIFF
--- a/_data/sw_bridges.yml
+++ b/_data/sw_bridges.yml
@@ -1,0 +1,128 @@
+- name: Bridges
+  software:
+    - name: BitlBee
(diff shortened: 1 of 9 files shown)


DarthGandalf:
Failing after 1m — Deploy failed

progval:
it's happening on master too, looks like the Jekyll version on Netlify is broken or something

DanielOaks:
I tried clearing the cache and re-deploying master in netlify and that deployed fine, probably a temp glitch on their side

PR #498, opened 2023-06-04
specify FAIL NICK NICKNAME_RESERVED

slingamn:
This is used in Ergo, e.g.

https://github.com/ergochat/ergo/blob/ca03a42dff12cc7dd3d622e1fd8dbb78447be810/irc/nickname.go#L49

In general Ergo sends standard replies without checking the standard-replies capability, but I'm associating this here with the standard-replies spec.


slingamn:
Looking at the rendered page, associating this with the standard-replies spec seems inappropriate.

emersion:
How does this related to `ERR_NICKLOCKED`?

progval:
and/or ERR_ERRONEUSNICKNAME

slingamn:
In Ergo, this is sent alongside `ERR_NICKNAMEINUSE`. I'm not familiar with how those other numerics are used in practice. (Are they sent by any major extant implementations?)

slingamn:
Sorry, correction, I am familiar with `ERR_ERRONEUSNICKNAME` and its uses. It seems inappropriate here because the client might conclude that the nickname is actually invalid, whereas typically it's valid (and could be claimed by the current client, given the right authentication).

slingamn:
FWIW Modern documents [ERR_NICKLOCKED](https://modern.ircdocs.horse/#errnicklocked-902) as having a different meaning:

>This numeric indicates that [SASL authentication](https://modern.ircdocs.horse/#authenticate-message) failed because the account is currently locked out, held, or otherwise administratively made unavailable. For more information on this numeric, see the IRCv3 [sasl-3.1](http://ircv3.net/specs/extensions/sasl-3.1.html) extension.

I didn't know about this because Ergo conflates this case with `ERR_SASLFAIL` [shrug]

slingamn:
Another reason to prefer `ERR_NICKNAMEINUSE` plus `FAIL` over `ERR_ERRONEUSNICKNAME`: like other cases of `ERR_NICKNAMEINUSE`, this condition can typically be fixed by growing a tail, which is not typically the case for nicknames that are "inherently" invalid (contain invalid characters, mixed scripts, etc.)

emersion:
Right, that makes sense to me!

jwheare:
Can we make this link to this PR as the spec.

See here for an example of how this is done (see pr-bot in _data/specs.yml) https://github.com/ircv3/ircv3.github.io/commit/f4089e8f119ffee5916132720190636e720a57a5

slingamn:
@jwheare I'm not sure that technique works anymore after 938e6c40402dfd4b07b4251c55668a4358d11df4 ; note for example that the resume numerics on [https://ircv3.net/registry](https://ircv3.net/registry) currently have a blank spec field (we should probably remove the resume numerics?)

jwheare:
OK, could we at least just add a comment linking here in the yaml.

PR #499, opened 2023-06-11
The Lounge and irc-framework do not support msgid (yet)

progval:
Grepping "msgid" in its source and irc-framework's doesn't return anything relevant.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -582,7 +582,6 @@
           extended-join:
           invite-notify:
           message-tags:
-          msgid:
           multi-prefix:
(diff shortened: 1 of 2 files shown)


ItsOnlyBinary:
irc-framework parses message-tags and add them to the emitted events, msgid would be included in the tags object of all events that include it.

progval:
@ItsOnlyBinary my rationale here is that it doesn't do anything with `msgid` specifically. For the same reason, it's not marked as supporting `+draft/react`, `+draft/reply`, +draft/channel-context`, or `+typing`

PR #500, opened 2023-06-26
Update WeeChat client support with version 4.0.0

flashcode:
Now supported in v4.0.0:
- batch
- echo-message
- multiline
- utf8only

Already supported: bot-mode.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -436,33 +436,33 @@
       support:
         stable:
           account-notify:
-          account-tag: 3.5+
+          account-tag:
           away-notify:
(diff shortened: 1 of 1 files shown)


progval:
what does weechat do with the bot mode?

flashcode:
@progval: nothing specific, but the message 335 is correctly handled and displayed, and the WHO bot flag should be just displayed as-is.

progval:
I see, thx

jwheare:
Thanks, can you also remove the version number from anything less than 3.6 (more than a year old)

flashcode:
@jwheare: done.

PR #521, opened 2024-07-19
Add account-extban spec

progval:
cc @causal-agent

### DIFF
--- a/_irc/index.md
+++ b/_irc/index.md
@@ -120,6 +120,10 @@ logged into accounts. This allows for much greater integration between client
 bots and the network's authentication system, as well as better general display
 and authentication of client identities.
 
+The [`account-extban` spec]({{site.baseurl}}/specs/extensions/account-extban.html)
+defines a way to either prevent a specific user account from joining a channel,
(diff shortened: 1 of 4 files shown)


_irc/index.md
```diff
@@ -120,6 +120,10 @@ logged into accounts. This allows for much greater integration between client
 bots and the network's authentication system, as well as better general display
 and authentication of client identities.
 
+The [`account-extban` spec]({{site.baseurl}}/specs/extensions/account-extban.html)
+defines a way to either prevent a specific user account from joining a channel,
+make it exempt from bans, or permanently allow it to an invite-only channel.
```
jwheare, line 125:
```suggestion
The [`account-extban` spec]({{site.baseurl}}/specs/extensions/account-extban.html) defines an ISUPPORT token allowing clients to construct an EXTBAN targeting a specific user account, e.g. to prevent them from joining a channel, make it exempt from bans, or permanently allow it to an invite-only channel.
```

Slightly more accurate description.

progval: I felt it was too technical for a summary, but ok

PR #527, opened 2024-08-04
clean up draft/resume registry entries

slingamn:


jwheare:
Is rendering of proposed specs even working? Might be worth just removing this whole section and associated validation rules until it's needed and can be displayed on the site in some way again.

slingamn:
I'm not sure, at least this way there is something there to debug? When we get it working again it'll probably look a lot like it does now.

PR #530, opened 2024-08-13
Update glirc support table

TheDaemoness:
Hi, collaborator on [glirc](https://github.com/glguy/irc-core) here. With the 2.41 release [right around the corner](https://github.com/glguy/irc-core/commit/82a318aadbdf9a6e7c46131845da2eebbad7905f), I figured now would be a good time to PR an update glirc's row in the client support table. This commit also indicates when support for various IRCv3 features was added for every version of glirc newer than the oldest packaged version (2.24 on Ubuntu 18.04), as well as adds two missing SASL mechanisms.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -120,30 +120,37 @@
           whox:
         SASL:
           - plain
-    - name: Glirc
-      # ref: https://github.com/glguy/irc-core/blob/v2/src/Client/State/Network.hs#L766
+    - name: glirc
+      # ref: selectCaps in https://github.com/glguy/irc-core/blob/v2/src/Client/State/Network.hs
       link: https://hackage.haskell.org/package/glirc
       support:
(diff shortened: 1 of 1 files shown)


_data/sw_clients.yml
```diff
@@ -120,30 +120,37 @@
...
+          sts: 2.30+
+          userhost-in-names: 2.30+
+          whox: 2.40+
         SASL:
           - plain
           - external
           - scram-sha-1
           - scram-sha-256
+          - ecdsa-nist256p-challenge
+          - ecdh-x25519-challenge
```
jwheare, line 153:
```suggestion
```
We don't include these on the site.

jwheare:
Version numbers aren't meant to be exhaustive project documentation. We try to make a sensible cut off of a couple of years back. Not a big deal, just for future reference.

PR #434, opened 2022-02-23
Bump submodule, add WHOX, reorganize groups in the spec list

progval:

PR #535, opened 2024-09-20
Update Sopel's support data for 8.0

dgw:
Filled in or corrected a few features added in older releases, too, e.g. the minimum version number for WHOX (added in #434).


dgw:
If you want I'll trim the older versions (before 7) in my next PR, after we ship 8.1.

For this one I was just mimicking what already existed in the file.

PR #548, opened 2025-01-05
Remove tldr.chat

alexpdp7:
This seems to be squatted by some AI product.


progval:
And according to `whois`, the domain changed ownership since it was submitted here.

PR #570, opened 2025-10-15
Solanum now supports message-tags and msgid

skizzerz:


progval:
That's the only kind of release Solanum does

SadieCat:
Maybe a date then?

PR #584, opened 2026-02-04
Add AndroidIRCx client with support details

r0073rr0r:


_data/sw_clients.yml
```diff
@@ -820,6 +820,46 @@
...
+          sts:
+          whox:
+          reply-client-tag:
+          channel-context-client-tag:
+          react-client-tag:
+          typing-client-tag:
+        SASL:
+          - external
+          - plain
+          - scram-sha-256
```
progval: SCRAM is not supported: https://github.com/AndroidIRCx/AndroidIRCx/blob/024566493ae951a92045f767781f3c1648db0805/src/services/IRCService.ts#L7450-L7494

progval (changes requested):

r0073rr0r:
Thanks for explanation ;)

PR #601, opened 2026-04-01
Add IRCafe to desktop clients

wodencafe:
I am the maintainer of [IRCafe](https://github.com/wodencafe/ircafe), and I wanted to submit it here to the list, as it supports many IRCv3 capabilities.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -249,6 +249,53 @@
           sts:
         SASL:
           - plain
+    - name: IRCafe
+      # ref: https://github.com/wodencafe/ircafe#ircv3-and-bouncer-support
+      # ref: https://github.com/wodencafe/ircafe/blob/main/src/main/java/cafe/woden/ircclient/irc/ircv3/Ircv3CapabilityCatalog.java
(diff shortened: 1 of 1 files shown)


jwheare:
Do you have screenshots of your support for replies/react/redaction/typing?

Looking at your [REQUESTABLE_CAPABILITIES](https://github.com/wodencafe/ircafe/blob/c231c23a174d29560aacc6335c69950d21872e1c/src/main/java/cafe/woden/ircclient/irc/ircv3/Ircv3CapabilityCatalog.java):

Several of these are tags and not capabilities, and I would recommend against requesting non-draft versions of capabilities that are still in draft, the final versions may have changes.

The `message-edit` capability is not even an approved draft, it's still a work in progress proposal that will likely be significantly reworked.

jwheare:
Closing until submitter is ready to resubmit.

wodencafe:
react
React: ^
redact
Redact: ^
reply
Reply: ^
typing
Typing: ^

@jwheare Sorry it took so long for me to get back to you on this, I've reworked this code significantly since then.

I think it's in good enough shape where it's ready for another review, if you're cool with that :sweat_smile: 
Thank you, and sorry for the issues with the earlier submission.

PR #617, opened 2026-06-04
Add ZoiteChat client

deepend-tildeclub:
Add Zoitechat client. A GTK3+ client based on Hexchat.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -577,10 +577,13 @@
           cap-3.1:
           cap-3.2:
           chghost:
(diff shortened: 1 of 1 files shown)


sadiepowell:
Now we have a successor on the table we can probably remove HexChat?

jwheare:
Thanks, can you remove scram-sha-1 and 512. I know some other old one list 1, but we try to keep this to the list of mechs that are actually listed https://deploy-preview-617--ircv3.netlify.app/docs/sasl-mechs

PR #618, opened 2026-06-16
Update the specs submodule, move clients that only implement +draft/channel-context to partial support.

sadiepowell:


jwheare:
channel-context is still labelled as a draft on:
https://deploy-preview-618--ircv3.netlify.app/irc/

The newly added oper-tag should get a mention on that page too, as well as in ‎_data/irc_versions.yml and _data/registry.yml

sadiepowell:
@jwheare Done & rebased against the current Git master.

PR #619, opened 2026-06-16
Update the site dependencies.

sadiepowell:
This was necessary for me to get the site to build on modern systems whilst working on #618 (Ubuntu 26.04 with RV 0.6.0).

---

- Update .ruby-version to the latest version of Ruby 3.

- Remove the ruby version from the validation workflow, the action script will use the version from .ruby_version here.

- Remove webrick from the Gemfile, this was for a bug in Jekyll on old versions but it has been fixed now.

- Fix a deprecation warning for site.name being renamed to site.title on recent Jekyll.

- Re-run Bundler with the recent changes.

- While I'm touching stuff update my email address in the mailmap.

- Add a netlify config file to build the site using Bundler.


jwheare:
Tried to remove a newline at the end via the web interface but the change didn't take :/

PR #624, opened 2026-07-08
bIRC Desktop Client

freshe:
Adds bIRC, a native macOS IRC client.
Supported extensions are documented at https://birc.app/ircv3.

### DIFF
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -62,6 +62,63 @@
           whox:
         SASL:
           - PLAIN
+    - name: bIRC
+      # ref: https://birc.app/ircv3
+      link: https://birc.app
+      support:
+        stable:
+          account-extban:
+          account-notify:
+          account-registration:
+          account-tag:
+          away-notify:
+          batch:
+          bot-mode:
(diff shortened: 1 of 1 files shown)


jwheare:
Thanks for the submission. Can you confirm whether this was built with AI assistance or not?

freshe:
Thanks for getting back. This is not a vibe-coded slop app - nor is it a compromised iOS/iPadOS or Electron cross-platform experience.

This is a true Mac native app, that took months of hard work to get out the door. 

And since Textual is now sunset, this is pretty much the only modern replacement. It already has a decent user base.

I have been developing Mac apps for many years.

https://www.fredrikblank.com

But to be totally transparent - I have used Claude to do a full bug audit of the codebase. And it actually found some narrow edge case bugs, that would have been hard to track down manually without feedback.

jwheare:
Thanks for the response. We're revising our approach a little for these submissions. Not to ban AI assisted apps entirely, but we are looking to require disclosure and an indication in the yaml/website in future.

freshe:
Many thanks!

PR #630, opened 2026-07-26
UnrealIRCd supports draft/multiline

ilmari-lauhakangas:
Implemented in 6.2.4, enabled by default in 6.2.6. See release notes: https://github.com/unrealircd/unrealircd/blob/68e5ce7cbba596a1ea02d77a15c70cbf8c43f118/doc/RELEASE-NOTES.md

### DIFF
--- a/_data/su_networks.yml
+++ b/_data/su_networks.yml
@@ -138,6 +138,7 @@
           monitor:
           msgid:
           multi-prefix:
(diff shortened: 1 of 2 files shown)


jwheare:
Thanks, can you add it to the networks file too (no need for versioning on that)

PR #633, opened 2026-07-26
Add Mango to the client tables

kanzelsberger:
Mango IRC is an IRCv3 capable client for iOS and macOS, available at https://mangoirc.chat


jwheare:
I will rely on other IRCv3 contributors to decide on whether this should be merged.

jwheare:
Thanks for being civil and bearing with us. Apologies for the curt initial dismissal.

Process note: submitter asserts that the app is not AI assisted, just the website.

PR #637, opened 2026-07-30
Allow marking software as using AI in their development.

sadiepowell:
Given the recent debates over AI use I think its reasonable to propose marking software so people reading the software lists can make a fully informed choice.

I've marked ObsidianIRC as I know for sure that is AI assisted but if there is any software I have missed please notify me and I will mark them.

We should probably also have some kind of rule for what actually consists AI assisted. I'd personally lean towards marking anything substantially AI assisted to avoid potential witch hunts about whether minor patches are AI or not.

---

This piggybacks off the existing support for labelling OS support. I can change this if necessary but it felt like I would just be duplicating code to do anything else.

### DIFF
--- a/_data/validation/sw_list.types.yaml
+++ b/_data/validation/sw_list.types.yaml
@@ -13,6 +13,9 @@ software:
           type: string
         link:
           type: string
+        ai:
+          type: boolean
+          required: false
         ircd-ver:
           type: string
           required: false
--- a/_data/sw_clients.yml
+++ b/_data/sw_clients.yml
@@ -780,6 +780,7 @@
       #ref https://github.com/ObsidianIRC/ObsidianIRC/blob/609af22ee1c30087e8d9125951f749501398617d/src/store/index.ts#L1035-L1083 (Typing)
       #ref https://github.com/ObsidianIRC/ObsidianIRC/blob/609af22ee1c30087e8d9125951f749501398617d/src/lib/ircClient.ts#L450 (SASL)
       link: https://github.com/ObsidianIRC/ObsidianIRC
+      ai: true
       support:
         stable:
           account-notify:
--- a/_includes/support_list.html
+++ b/_includes/support_list.html
@@ -70,8 +70,11 @@
                 {% if sw.note %}
                     <div class="sw-note">{{ sw.note | markdownify }}</div>
                 {% endif %}
-                {% if sw.os %}
+                {% if sw.os or sw.ai %}
                     <div class="os-support">
+                    {% if sw.ai %}
+                         <i title="Development assisted using AI" class="fa-solid fa-robot os-support-ai"></i>
+                    {% endif %}
                     {% for name in sw.os %}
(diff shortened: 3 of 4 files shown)


jwheare:
Thanks, I think the icon can be less emphasised colour-wise: #bbb maybe

jwheare:
Keeping it in sw.ai is fine, but class name could be changed from `os-support-ai` to `ai-assisted`

jwheare:
LGTM

_data/validation/sw_list.types.yaml
```diff
@@ -13,6 +13,9 @@ software:
           type: string
         link:
           type: string
+        ai:
+          type: boolean
+          required: false
```
Celelibi, line 18: I think it's at least very restrictive to make it a boolean. At what point does it become true?
- Exchanging ideas with a chat bot about creating said-software before any file has been created?
- Using a chat bot for early designing?
- Using it for debugging?
- Using it for refactoring?
- Using it as a smarter text-completion in an IDE?
- Generating parts of code in a chat bot that they copy-paste with or without reading it?
- Using an agent but spending more time reviewing the code than you'd have spent writing it?
- Using an agent but giving the review a varying amount of effort?
- Using an agent to generate commit messages, but not the code?
- Using an agent to generate the code but not the commit messages?
(I could keep the list going on forever, but you get the idea.)

And when there are multiple contributors, each will probably have a different degree of assistance and a different way of using it.
If inclusion of this information is unavoidable (I'd rather avoid it, as I think it's not very informative IMO), it should at least allow for more nuance.

jwheare: I don't think we need to be the keepers of this nuance. It's a hint, readers can make their own investigations.

Celelibi: Then we need to at least decide an arbitrary threshold.
And still, it might not be entirely known by the maintainer and change through time.

I, for instance, regularly use ChatGPT to bounce ideas, to refine the scope of the project, to prevent scope creep, to find good reasons not to implement some features, to find cleverer ways to implement others, to iterate on designs decisions, to discuss apparent contradictions, and much more.
I also sometimes use Claude to generate code. But when I do, I retype everything manually to make sure I understand every line and every interaction with every other line.

Would that award me an AI-star of death?

I personally don't think I should have to disclose where my ideas come from... assuming I even know. My ideas and my code are ultimately mine and should be judged as such.

jwheare: I think the line is probably relatively obvious on a case by case basis, but hard to enumerate and write hard and fast guidelines that won't go out of date (as you mentioned) or be rules lawyered into oblivion.

At the end of the day, these lists don't aspire to any accuracy or completeneess guarantees, and they are ultimately lightly curated. It will be a subjective call made by the maintainers, using their best judgment, but it will also be subject to scrutiny and revision. There is not much need to over-dramatise.

PR #640, opened 2026-08-04
Update spec support for Teranova.

sadiepowell:


jwheare:
What does "known client tags" mean? Still looks like `CLIENTTAGDENY=*` i being sent.

sadiepowell:
> What does "known client tags" mean? Still looks like `CLIENTTAGDENY=*` i being sent.

Oops, you're on `sigsegv.*` which hasn't been updated yet. It should be updated in a few hours when the sysadmin wakes up and pulls our config repo. I wasn't expecting you to get around to checking it so fast. 😅

Known means filtered to client tags that are in the IRCv3 tag registry (our compromise to get client tag-sceptical admins to enable client tags). I can change it to something else if you'd prefer it?

jwheare:
No strong preference on wording. Libera says "Filtered client tags" but I think that's more restricted. Probably should get networks to indicate which ircv3 client tags they do support nowadays.

